PR #2, opened 2022-02-10
add crs templates and files for AWS

calvix:
create a template for CRs which includes defaulting from kyverno

build based on https://github.com/giantswarm/cluster-openstack/


calvix:
anyone anyhting?

helm/cluster-aws/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+machinePools:
+- name: def00  # Name of node pool.
+  availabilityZones:
+  - eu-west-1
+  instanceType: m5.xlarge
+  minSize: 3  # Number of replicas in node pool.
+  maxSize: 3
+  rootVolumeSizeGB: 300
+  customNodeLabels:
+  - label=default
```
fiunchinho, line 37: will this label be the one used to select the nodepool for the workloads?

calvix: this is just example of a custom label, you still get the  one we set on our current product `giantswarm.io/machine-pool=$MACHINE_POOL_ID`
 https://github.com/giantswarm/cluster-aws/pull/2/files#diff-bbbbf26ad998e5d32de273c75850a577a47216033b53d5d6f2799d0a30d4028fR75

fiunchinho (approved):
Let's move forward, we can always iterate later

njuettner (approved):
let's go 🚀

helm/cluster-aws/files/opt/init-disks.sh
```diff
@@ -0,0 +1,30 @@
...
+	KUBELET_DISK="/dev/nvme3n1"
+	echo found nvme DISKs
+else
+	ETCD_DISK="/dev/xvdc"
+        CONTAINERD_DISK="/dev/xvdd"
+        KUBELET_DISK="/dev/xvde"
+fi
+
+# init etcd disk
+mkfs.xfs $ETCD_DISK
```
whites11, line 15: can this run twice? will it erase disk in an undesirable way?

whites11 (approved):
I literally have no idea what this PR is about, but LGTM

helm/cluster-aws/files/opt/init-disks.sh
calvix: In CAPI the machines do not stay, so during an update new machine is created and old removed

there might be some race condition but whole CAPI on ubuntu and cloud init is disaster to happen an di hope we can fix this on ignition

PR #17, opened 2022-03-04
remove-cluster-role-identity-from-repo

calvix:
It does not seem to be a very good idea to have AWSClusterRole as part of this app, as deletion of it will cause cluster stuck in deletion

instead, we will create a default role with capa controller app


AverageMarcus:
How about leveraging Helm pre/post hooks to handle the creation and deletion of this resource?

calvix:
> How about leveraging Helm pre/post hooks to handle the creation and deletion of this resource?

How would that help?

AverageMarcus:
Ensures the `AWSClusterRole` isn't removed until the cluster is.

calvix:
> Ensures the `AWSClusterRole` isn't removed until the cluster is.

In the end, I think having a unique `AWSClusterRole` for every cluster is not such a good idea, and rather have a similar way how we to credentials now, it's only one unique ARN role, and than you reference that role in the CR via name. using `default` as default

It also simplifies the required arguments you need in order to generate yamls for a cluster.

AverageMarcus:
Makes sense. 👍

AverageMarcus (approved):

PR #20, opened 2022-03-07
convert-aws-account-to-string

calvix:
```
    reason: 'Upgrade "vacl7" failed: failed to create resource: AWSMachineTemplate.infrastructure.cluster.x-k8s.io
      "vacl7-bastion" is invalid: spec.template.spec.imageLookupOrg: Invalid value:
      "integer": spec.template.spec.imageLookupOrg in body must be of type string:
      "integer"'
```


calvix:
just merging

PR #33, opened 2022-03-15
set-max-db-size

calvix:
set etcd max db size to 8gb to avoid issues we see at one of our customers


AverageMarcus (approved):
LGTM

I assume the nodes have enough memory available?

calvix:
the size itself is more about the maximum allowed size on the disk, it might affect memory consumption but only on clusters that already have gigantic machines so IMHO RAM should not be issue if we increase this value

PR #39, opened 2022-03-21
rename networkSpec to network

calvix:
when those rules is missing in the CRS, capa will empty the rules and add them,  causing removal of the rules we set for was cni


AverageMarcus (approved):
LGTM

Changelog need updating?

PR #46, opened 2022-03-30
Lookup region and AZs from management cluster

AverageMarcus:
* If the AWS region isn't set in the variables we'll attempt to pull it from the management cluster
* If the availability zones on a node pool aren't set we'll pull them from the management cluster

Towards: https://github.com/giantswarm/giantswarm/issues/21298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Lookup AWS region if not set in values
+- Lookup AWS Availability Zones if not set in values
 
 ## [0.2.0] - 2022-03-29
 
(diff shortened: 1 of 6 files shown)


whites11 (approved):
dude, this is fucking smart!

fiunchinho:
why do we want these to be optional values? I think it's normal to have to select region and AZs when creating a cluster.

I also find it weird to default the AZs to whatever we decided for the management cluster.

I'm not blocking this, anyway.

fiunchinho (approved):
I did not know this was possible, nice.

AverageMarcus:
Why optional? - for region it makes sense as we (at least so far) have enforced WCs be created in the same region as the MC so is a bit pointless having to specify it every time you do `kubectl gs template cluster ....`

With the AZs, defaulting them to the same as the MC makes sense for most general cases (e.g. in eu-west-1 there's only 3 AZs to pick from so if the cluster is HA, which is is by default, then you'll always be selecting all three AZs) For the cases where different AZs are needed, those can still be provided (either on the template directly, or via a soon to be optional flag on `kubectl gs template`)

AverageMarcus:
Related kubectl-gs change: https://github.com/giantswarm/kubectl-gs/pull/727

PR #53, opened 2022-04-14
Updated to use latest cluster-shared library chart

AverageMarcus:
Updated to use the latest https://github.com/giantswarm/cluster-shared as a library chart.

We are also explicitly setting the values the library will use (even though they are same as the defaults) to make it clear what values are used.

Added support for providing a custom `clusterName` value which defaults to the release name if not provided (as with the pre-existing behaviour)


njuettner (approved):
lgtm

PR #56, opened 2022-04-28
Add values.schema.json file

fiunchinho:


whites11 (approved):
If you copy-pasted it than it's ok

PR #65, opened 2022-07-26
Set pod CIDR range

AverageMarcus:
Set pod CIDR to 100.64.0.0/12 to match what we set in Cilium (and to not clash with AWS CIDR) and add needed flags to controller-manager


mnitchev (approved):
LGTM

PR #77, opened 2022-08-08
Switch to using GS-built AMIs

AverageMarcus:
Fixes https://github.com/giantswarm/roadmap/issues/1290

Makes use of the CAPA AMIs that we build ourselves.

/test create

### DIFF
--- a/Makefile.custom.mk
+++ b/Makefile.custom.mk
@@ -1,3 +1,9 @@
+.PHONY: template
+template:
+	@cd helm/cluster-aws && \
+		sed -i '' 's/version: \[/version: 1 #\[/' Chart.yaml && \
+		helm template . && \
+		sed -i '' 's/version: 1 #\[/version: \[/' Chart.yaml
 
 ensure-schema-gen:
 	@helm schema-gen --help &>/dev/null || helm plugin install https://github.com/mihaisee/helm-schema-gen.git
--- a/helm/cluster-aws/templates/_helpers.tpl
+++ b/helm/cluster-aws/templates/_helpers.tpl
@@ -90,3 +90,15 @@ room for such suffix.
 {{- define "bastionIgnition" }}
 {{- tpl ($.Files.Get "files/bastion.iqn") . | b64enc}}
 {{- end -}}
+
+{{- define "ami" }}
+{{- if .Values.ami }}
+ami:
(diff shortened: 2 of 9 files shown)


fiunchinho (approved):

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -90,3 +90,15 @@ room for such suffix.
...
+
+{{- define "ami" }}
+{{- if .Values.ami }}
+ami:
+  id: {{ .Values.ami }}
+{{- else -}}
+ami: {}
+imageLookupBaseOS: "ubuntu"
+imageLookupFormat: {{ "capa-ami-{{.BaseOS}}-{{.K8sVersion}}-00-gs" }}
+imageLookupOrg: "706635527432"
```
fiunchinho, line 102: where are these used now?

Makefile.custom.mk
```diff
@@ -1,3 +1,9 @@
+.PHONY: template
+template:
+	@cd helm/cluster-aws && \
+		sed -i '' 's/version: \[/version: 1 #\[/' Chart.yaml && \
+		helm template . && \
+		sed -i '' 's/version: 1 #\[/version: \[/' Chart.yaml
```
fiunchinho, line 6: [On other projects](https://github.com/giantswarm/dns-operator-gcp/blob/main/Makefile.custom.mk#L132) we use `architect helm template` for this. Maybe it would be better to align

AverageMarcus: I tried that. It didn't work on my machine. Tried to fix it but the scary define block was beyond me so I went "simple" 😆

helm/cluster-aws/templates/_helpers.tpl
AverageMarcus: * https://github.com/giantswarm/cluster-aws/pull/77/files#diff-dee2f6d641136e8abf8d1eb197ba9edd300dc861cf4cde901204c6b890f41dd7R118
* https://github.com/giantswarm/cluster-aws/pull/77/files#diff-bbbbf26ad998e5d32de273c75850a577a47216033b53d5d6f2799d0a30d4028fR41

fiunchinho (approved):

PR #88, opened 2022-08-12
Added hash function for immutable resource name suffix

AverageMarcus:
Fixes https://github.com/giantswarm/roadmap/issues/1308


fiunchinho (approved):
I guess you checked the `MachinePools` and the hash is not needed with those?

AverageMarcus:
Correct, it's just the templates that need it. (Plus some resources we don't use like the ones for EKS and Fargate)

PR #94, opened 2022-08-16
Added AZ filter to machinepool subnet filters

AverageMarcus:
Signed-off-by: Marcus Noble <github@marcusnoble.co.uk>

### DIFF
--- a/helm/cluster-aws/templates/_machine_pools.tpl
+++ b/helm/cluster-aws/templates/_machine_pools.tpl
@@ -45,6 +45,8 @@ spec:
     - name: tag:sigs.k8s.io/cluster-api-provider-aws/role
       values:
       - private
+    - name: availabilityZone
+      values: {{ .availabilityZones | default ( include "aws-availability-zones" .) }}
   awsLaunchTemplate:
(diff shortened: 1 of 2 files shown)


helm/cluster-aws/templates/_machine_pools.tpl
```diff
@@ -45,6 +45,8 @@ spec:
     - name: tag:sigs.k8s.io/cluster-api-provider-aws/role
       values:
       - private
+    - name: availabilityZone
+      values: {{ .availabilityZones | default ( include "aws-availability-zones" .) }}
```
calvix, line 49: this is not a tag but a property?

calvix: does it properly template with one or multiple AZs?

mnitchev (approved):
Don't have much context on aws filters, but looks good otherwise.

helm/cluster-aws/templates/_machine_pools.tpl
AverageMarcus: Yes, to both.

PR #96, opened 2022-08-17
Added network topology annotations

AverageMarcus:
Initial work towards https://github.com/giantswarm/roadmap/issues/1306

Adds a value to toggle the "network topology" that the cluster will use. This will default to `GiantSwarmManaged` and but the initial focus for implementation. A `transitGatewayID` value is also added for when `UserManaged` is set.

The `network-topology.giantswarm.io/*` annotations will be used by a (currently non-existent) `transit-gateway-operator`.

### DIFF
--- a/helm/cluster-aws/values.yaml
+++ b/helm/cluster-aws/values.yaml
@@ -15,6 +15,12 @@ network:
   serviceCIDR: 172.31.0.0/16
   podCIDR: 100.64.0.0/12
 
+  # topologyMode defines the type of cross-cluster networking architecture between MCs and WCs
+  # Valid values: GiantSwarmManaged, UserManaged, None
+  topologyMode: GiantSwarmManaged
+  # transitGatewayID is the ID of the TGW to use when `mode` is set to `UserManaged`.
+  transitGatewayID:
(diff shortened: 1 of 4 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -15,6 +15,12 @@ network:
   serviceCIDR: 172.31.0.0/16
   podCIDR: 100.64.0.0/12
 
+  # topologyMode defines the type of cross-cluster networking architecture between MCs and WCs
+  # Valid values: GiantSwarmManaged, UserManaged, None
+  topologyMode: GiantSwarmManaged
+  # transitGatewayID is the ID of the TGW to use when `mode` is set to `UserManaged`.
+  transitGatewayID:
```
calvix, line 22: IMHO `transitGatewayID: ""` would be better so the type is a string in the values.schema

calvix (approved):

helm/cluster-aws/values.yaml
AverageMarcus: Yeah, I agree. Will update

PR #99, opened 2022-08-18
add-ssh-rules

calvix:
configure ssh to listen on port 30000  on bastion node

few more tweaks nothing serious

### DIFF
--- a/helm/cluster-aws/files/bastion.iqn
+++ b/helm/cluster-aws/files/bastion.iqn
@@ -7,7 +7,7 @@
          {
             "name":"giantswarm",
             "sshAuthorizedKeys":[
-               "ssh-rsa AAAABEKf fake@giantswarm"
+               "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKLSRVtP/b9bcMPYOa49/rj+09bb9TP8L3kCyh4miDkr fake@giantswarm"
             ],
             "groups":[
                "sudo",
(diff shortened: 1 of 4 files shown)


helm/cluster-aws/files/bastion.iqn
```diff
@@ -7,7 +7,7 @@
          {
             "name":"giantswarm",
             "sshAuthorizedKeys":[
-               "ssh-rsa AAAABEKf fake@giantswarm"
+               "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKLSRVtP/b9bcMPYOa49/rj+09bb9TP8L3kCyh4miDkr fake@giantswarm"
```
calvix, line 10: the invalid key was causing some error noise during boot so giving this random non used public key

helm/cluster-aws/templates/_bastion.tpl
```diff
@@ -7,6 +7,8 @@ metadata:
     {{- include "labels.common" $ | nindent 4 }}
   name: {{ include "resource.default.name" $ }}-bastion-ignition
   namespace: {{ .Release.Namespace }}
+  finalizers:
+  - bastion.finalizers.giantswarm.io/secret-blocker
```
fiunchinho, line 11: how come we need this?

calvix: there is a bug/feature in CAPA controller that if the secret file that is referenced in bootstrap data is gone, the reconciliation will fail and the awsmachine is not deleted(it will error on no bootstrap data found for some reason), this will go away once we migrate to flatcar as we no longer use custom ignition but it will be generated and managed by the bootstrap controller

calvix: this was an issue before but appeared few version again

PR #103, opened 2022-08-24
allow-single-cp

calvix:
silently allow setting control-plane replicas


calvix:
I remember @AverageMarcus you removed this option, do you think its fine to have it hidden?

This would be for the proxy cluster in order to make ti small as possible

fiunchinho (approved):

AverageMarcus:
Yeah, I think this is fine to have. We (Giant Swarm generally) might need to think about how we handle situations where these values are "discovered" and abused though. Maybe something for the SAs.

PR #110, opened 2022-09-06
fix-proxy-file-name

calvix:
fix proxy file name and the helm context
fix control plane template spec to use proper function


helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -62,7 +62,7 @@ room for such suffix.
 - path: /etc/systemd/system/containerd.service.d/http-proxy.conf
   permissions: "0644"
   encoding: base64
-  content: {{ tpl ($.Files.Get "files/etc/systemd/system/containerd.service.d/http-proxy.conf") . | b64enc }}
```
AverageMarcus, line 65: Does `$` work here?

AverageMarcus (approved):

helm/cluster-aws/templates/_helpers.tpl
calvix: the template rendered fine and the nodes have proper proxy config so I assume yes

PR #121, opened 2022-10-11
Irsa for CAPA

njuettner:
Issue: https://github.com/giantswarm/roadmap/issues/1095

### DIFF
--- a/helm/cluster-aws/templates/_helpers.tpl
+++ b/helm/cluster-aws/templates/_helpers.tpl
@@ -77,6 +77,12 @@ room for such suffix.
     secret:
       name: {{ include "resource.default.name" $ }}-encryption-provider-config
       key: encryption
+- path: /etc/kubernetes/irsa/cloudfront.yaml
+  permissions: "0600"
+  contentFrom:
+    secret:
+      name: {{ include "resource.default.name" $ }}-irsa-cloudfront
+      key: domain
 {{- end -}}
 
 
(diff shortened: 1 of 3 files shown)


helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -88,6 +94,10 @@ room for such suffix.
 - /bin/sh /opt/init-disks.sh
 {{- end -}}
 
+{{- define "irsaPreKubeadmCommands" -}}
+- /bin/sh /opt/irsa-cloudfront.sh
```
calvix, line 98: this still need to be put into KubeAdmCOntrolplane CR

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -77,6 +77,12 @@ room for such suffix.
     secret:
       name: {{ include "resource.default.name" $ }}-encryption-provider-config
       key: encryption
+- path: /etc/kubernetes/irsa/cloudfront-domain
+  permissions: "0600"
+  contentFrom:
+    secret:
+      name: {{ include "resource.default.name" $ }}-irsa-cloudfront
+      key: domain
 {{- end -}}
```
calvix, line 86: you still miss the bash script  here, it needs to be also defined as a file with the content

ie like this https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/templates/_helpers.tpl#L55-L60

and than put that function here https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/templates/_control_plane.tpl#L113

calvix (approved):
lgtm

PR #122, opened 2022-10-11
Pause cluster is vpcMode is private

AverageMarcus:
Towards https://github.com/giantswarm/roadmap/issues/1289

Set the `Cluster` and `AWSCluster` to paused initially when `vpcMode` is set to `private`.

### DIFF
--- a/helm/cluster-aws/values.yaml
+++ b/helm/cluster-aws/values.yaml
@@ -18,6 +18,7 @@ network:
   # vpcMode defines if the VPC is created with public, internet facing resources (public subnets, NAT gateway) or not
   # Valid value: public, private
   vpcMode: public
+
   # apiMode defines if the Kubernetes API server load balancer should be public (internet facing) or private (internal only)
   # Valid value: public, private
   apiMode: public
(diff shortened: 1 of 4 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -18,6 +18,7 @@ network:
   # vpcMode defines if the VPC is created with public, internet facing resources (public subnets, NAT gateway) or not
```
fiunchinho, line 18: Should we explain here that the cluster would be paused when selecting `private`?

fiunchinho (approved):

helm/cluster-aws/values.yaml
AverageMarcus: It'll only be paused until the vpc operator finishes its work and unpauses. Do you think that needs to be called out? The timeframe of it being paused _should_ be fairly small.

fiunchinho: I guess it's an implementation detail and we can leave it out.

PR #136, opened 2022-10-20
Add `api-audiences` for IRSA

njuettner:

Audience was missing for IRSA, when requesting IAM roles audience has to be present when using the service account token.

### Checklist

### Trigger e2e tests

/test create
/test upgrade


AverageMarcus (approved):
LGTM

I assume we'll always have IRSA in our CAPA clusters, yes?

njuettner:
@AverageMarcus correct 👍🏻

PR #149, opened 2022-11-04
Support setting custom node taints

AverageMarcus:
Signed-off-by: Marcus Noble <github@marcusnoble.co.uk>

Fixes https://github.com/giantswarm/roadmap/issues/1586

Supports setting custom taints on both worker nodes and control plane nodes. 

Note: does not allow for removing of default taints! (by design)

### Checklist

### Trigger e2e tests

No E2E cluster available yet :(

### DIFF
--- a/helm/cluster-aws/values.yaml
+++ b/helm/cluster-aws/values.yaml
@@ -60,6 +60,10 @@ machinePools:
   rootVolumeSizeGB: 300
   customNodeLabels:
   - label=default
+  customNodeTaints: []
+  # - key: ""
+  #   value: ""
+  #   effect: "" # Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
 
 sshSSOPublicKey: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIM4cvZ01fLmO9cJbWUj7sfF+NhECgy+Cl0bazSrZX7sU vault-ca@vault.operations.giantswarm.io"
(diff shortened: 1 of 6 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -50,6 +50,10 @@ controlPlane:
   etcdVolumeSizeGB: 100
   containerdVolumeSizeGB: 100
   kubeletVolumeSizeGB: 100
+  customNodeTaints: []
+  # - key: ""
+  #   value: ""
+  #   effect: "" # Valid effects are NoSchedule, PreferNoSchedule and NoExecute.
```
calvix, line 56: i don't think we should allow customer to set custom taints on masters, that seems dangerous, they should not run stuff on masters

calvix (approved):
lgtm but not sure about taints on master/control-plane node, maybe something for discussion in team

helm/cluster-aws/values.yaml
AverageMarcus: Yeah I wasn't too sure either. 

How about a half-way and have the logic in the templates but don't explicitly mention it in the values file?

calvix: ok that sounds better, so its possible to do it but customer wont find it easy , works for me

AverageMarcus: Updated :)

calvix (approved):

PR #160, opened 2022-11-18
add-containerd-registry-auth

calvix:
add option to configure docker auth

### DIFF
--- a/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
+++ b/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
@@ -7,4 +7,4 @@ endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
 [plugins."io.containerd.grpc.v1.cri".registry.configs]
 [plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]
 username = "{{ .Values.registry.username }}"
-password = "{{ .Values.registry.password }}
+password = "{{ .Values.registry.password }}"
--- a/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
+++ b/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
@@ -1,0 +1,10 @@
+[plugins."io.containerd.grpc.v1.cri".registry]
+
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors]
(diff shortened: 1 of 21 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -88,6 +88,11 @@ proxy:
   http_proxy: ""
   https_proxy: ""
 
+registry:
```
AverageMarcus, line 91: Can we leave this hidden from the values? I'm still very much of the opinion that customers shouldn't need to configure this.

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -0,0 +1,10 @@
+[plugins."io.containerd.grpc.v1.cri".registry]
+
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors]
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
+endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
+
+[plugins."io.containerd.grpc.v1.cri".registry.configs]
+[plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]
```
AverageMarcus, line 8: Do we need auth for the azureecr reg also?

calvix: not sure about azure limits TBH but never seen that we hit those

helm/cluster-aws/values.yaml
fiunchinho: Yes, I think so. [According to the docs, the `.Values` object](https://helm.sh/docs/chart_template_guide/values_files/) 

> This object provides access to values passed into the chart. Its contents come from multiple sources:
> 
> - The values.yaml file in the chart
> - If this is a subchart, the values.yaml file of a parent chart
> - A values file if passed into helm install or helm upgrade with the -f flag (helm install -f myvals.yaml ./mychart)
> - Individual parameters passed with --set (such as helm install --set foo=bar ./mychart)

Meaning that when passing the value while installing/upgrading the chart, it will be available in `.Values`, even though is not defined (better say _defaulted_?) in the `values.yaml` file.

AverageMarcus: We could handle it like with the AMI - https://github.com/giantswarm/cluster-aws/blob/1691c83d4cf32ff4303de31d6db3e36118c73cc0/helm/cluster-aws/templates/_helpers.tpl#L139

PR #161, opened 2022-11-21
Make 'baseDomain' a required value

fiunchinho:

There was a problem setting the `baseDomain` last week, and clusters kept being created relying on the default `baseDomain`, which was set to example.com. This made that the clusters were not being created successfully.

Instead, this value should be always be passed and fail if that's not the case.

I also made `organization` required to align with `cluster-gcp`.

### Checklist

### Trigger e2e tests

/test create
/test upgrade


zewolfe (approved):
👍🏾

PR #162, opened 2022-11-22
Fix service account issuer

njuettner:
We currently facing normalizing issuer errors when trying to use IRSA. This is because I missed the "https://" when setting the `--service-account-issuer` API flag, it needs to start with "https://" otherwise it's not working.

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


fiunchinho (changes requested):
CHANGELOG :)

fiunchinho (approved):

fiunchinho:
Tests are currently not working for CAPA. Just ignore them.

njuettner:
Thanks <3

PR #166, opened 2022-11-22
enable external resource gc

bdehri:
### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/helm/cluster-aws/templates/_aws_cluster.tpl
+++ b/helm/cluster-aws/templates/_aws_cluster.tpl
@@ -11,7 +11,7 @@ metadata:
     {{- if (eq .Values.network.vpcMode "private") }}
     cluster.x-k8s.io/paused: "true"
     {{end}}
-    aws.cluster.x-k8s.io/external-resource-gc: "{{ .Values.aws.externalResourceGc }}"
+    aws.cluster.x-k8s.io/external-resource-gc: "true"
   labels:
     {{- include "labels.common" $ | nindent 4 }}
   name: {{ include "resource.default.name" $ }}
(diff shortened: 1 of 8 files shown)


fiunchinho (commented):
Is there a reason not to enable this? I'd always enable it and don't add it to the helm values. What do you think?

helm/cluster-aws/values.yaml
```diff
@@ -8,6 +8,7 @@ baseDomain: "example.com"
 aws:
   region: ""
   awsClusterRole: "default"
+  externalResouceGc: true
```
calvix, line 11: typo

calvix: can we add commetn above the value to explain what it does?

helm/cluster-aws/templates/_aws_cluster.tpl
```diff
@@ -11,6 +11,7 @@ metadata:
     {{- if (eq .Values.network.vpcMode "private") }}
     cluster.x-k8s.io/paused: "true"
     {{end}}
+    aws.cluster.x-k8s.io/external-resource-gc: "{{ .Values.aws.externalResourceGc }}"
```
calvix, line 14: where is this annotation came from?

bdehri:
> Is there a reason not to enable this? I'd always enable it and don't add it to the helm values. What do you think?

No specific reason actually, it is a little bit verbose in this way IMHO

calvix (approved):

calvix:
> Is there a reason not to enable this? I'd always enable it and don't add it to the helm values. What do you think?

yeah true, we always want to clean external resources, so maybe add it to the annotations, and no need to add it as a configurable value :p at least we can avoid a situation where someone disables it without thinking

bdehri:
@fiunchinho I saw you comment about these failing tests are ignorable. Is this still the case?

fiunchinho:
> @fiunchinho I saw your comment about these failing tests are ignorable. Is this still the case?

There were broken yesterday but I just fixed them. Let's try!

/test create
/test upgrade

PR #171, opened 2022-11-29
fix IRSA placeholder replacer

bdehri:

At the moment, CAPA cluster upgrades require manuel intervention due to service-account-issuer placeholder is not replaced on non-init nodes. This PR enables replacing of the placeholder on non-init nodes.
### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
+++ b/helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Replace Kubeadm Cloudfront placeholder
+CLOUDFRONT_DOMAIN="https://$(cat /etc/kubernetes/irsa/cloudfront-domain)"
+
+sed -i "s|PLACEHOLDER_CLOUDFRONT_DOMAIN|$CLOUDFRONT_DOMAIN|g" /etc/kubernetes/manifests/kube-apiserver.yaml
--- a/helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
+++ b/helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
@@ -1,6 +1,0 @@
-#!/bin/bash
-
-# Replace Kubeadm Cloudfront placeholder
-CLOUDFRONT_DOMAIN="https://$(cat /etc/kubernetes/irsa/cloudfront-domain)"
-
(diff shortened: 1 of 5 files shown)


calvix:
What if we just have one bash script for both and pass the file that should be replaced as an argument?

Also, we should put the sed command into `IF` and only execute in case the file exists to avoid unnecessary error messages that might confuse someone else debugging something later

helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Replace Kubeadm Cloudfront placeholder
+CLOUDFRONT_DOMAIN="https://$(cat /etc/kubernetes/irsa/cloudfront-domain)"
+
+sed -i "s|PLACEHOLDER_CLOUDFRONT_DOMAIN|$CLOUDFRONT_DOMAIN|g" /etc/kubernetes/manifests/kube-apiserver.yaml
```
AverageMarcus, line 6: I feel like I might have asked this before but can't remember the answer...

Is there any reason that patching this file can't also work for the initial creation? Do we need to patch the kubadm config on init or can we just rely on patching the api-server?

bdehri: Obviously, I do not know about the previous reasoning, but I can give it a try.

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -122,10 +122,14 @@ room for such suffix.
 - /bin/sh /opt/init-disks.sh
 {{- end -}}
 
-{{- define "irsaPreKubeadmCommands" -}}
+{{- define "irsaReplacerCommands" -}}
 - /bin/sh /opt/irsa-cloudfront.sh
 {{- end -}}
 
+{{- define "irsaPostKubeadmCommands" -}}
+- /bin/sh /opt/irsa-cloudfront-postkubeadm.sh
```
calvix, line 130: I thnik the argument should be here instead of https://github.com/giantswarm/cluster-aws/pull/171/files#diff-dee2f6d641136e8abf8d1eb197ba9edd300dc861cf4cde901204c6b890f41dd7R175

helm/cluster-aws/files/opt/irsa-cloudfront-postkubeadm.sh
bdehri: It works actually.

PR #173, opened 2022-11-30
fix IRSA placeholder replacer

bdehri:

At the moment, CAPA cluster upgrades require manuel intervention due to service-account-issuer placeholder is not replaced on non-init nodes. This PR enables replacing of the placeholder on non-init nodes.

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


AverageMarcus (approved):
Can I assume that only replacing in the api-server manifest works for both init and join then?

If so, LGTM :)

bdehri:
> Can I assume that only replacing in the api-server manifest works for both init and join then?
> 
> If so, LGTM :)

Yes

fiunchinho (approved):

PR #175, opened 2022-11-30
fix IRSA when upgrading clusters

njuettner:

When clusters being upgraded they are currently broken. We need to adjust the kubeadm config so that when new nodes are joining they won't use CLOUDFRONT_PLACEHOLDER_DOMAIN anymore.

### Checklist

### Trigger e2e tests

/test create
/test upgrade


njuettner:
Sorry I haven't seen this PR https://github.com/giantswarm/cluster-aws/pull/173, it might also work but I'd like to point out we would still miss `kubeadm-config` config map which still has the PLACEHOLDER.

helm/cluster-aws/files/opt/post-irsa-cloudfront.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Replace Kubeadm Cloudfront placeholder
+CLOUDFRONT_DOMAIN="https://$(cat /etc/kubernetes/irsa/cloudfront-domain)"
+
+# We need to store cloudfront domain in kubeadmconfig config map, because kubeadm join can't inject cloudfront domain 
+kubectl --kubeconfig /etc/kubernetes/admin.conf -n kube-system get cm kubeadm-config -o yaml > /tmp/kubeadmconfig.yaml
```
calvix, line 7: AFAIK this is periodically updated by clusterAPI, so this is change will not last, or I am wrong?

calvix:
also there seem to be some miscomunication as two people are working on same issue
https://github.com/giantswarm/cluster-aws/pull/173

helm/cluster-aws/files/opt/post-irsa-cloudfront.sh
njuettner: I just ran it multiple hours, the configmap was not overwritten again 🤔

calvix: Maybe I am thinking of something else, we we are certain that the change will stays even after update than I am fine.

njuettner:
We decided to continue using Berk's approach 👍🏻

PR #178, opened 2022-12-01
fix-container-registry

calvix:
workaround for the registry issue, tested on golem with 300 pulls, seems to working

### DIFF
--- a/helm/cluster-aws/templates/_helpers.tpl
+++ b/helm/cluster-aws/templates/_helpers.tpl
@@ -105,7 +105,7 @@ room for such suffix.
 {{- end -}}
 {{- end -}}
 
-{{- define registryWorkaroundCommands -}}
+{{- define "registryWorkaroundCommands" -}}
 - mkdir -p /etc/containerd/conf.d/
 - mv /opt/registry-config.toml /etc/containerd/conf.d/registry-config.toml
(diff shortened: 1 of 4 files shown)


tuladhar (approved):

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -105,6 +105,13 @@ room for such suffix.
 {{- end -}}
 {{- end -}}
 
+{{- define "registryWorkaroundCommands" -}}
```
QuentinBisson, line 108: Does it make sense to document why this workaround is actually needed so we do not forget?

calvix: good point

tuladhar: @QuentinBisson It's documented here: https://github.com/giantswarm/roadmap/issues/1737

calvix: i added comment line as well

QuentinBisson: Sure but I meant in this PR :) It can be a simple comment to link the roadmap issue ;)

QuentinBisson: Loving it  @calvix :)

PR #184, opened 2022-12-07
enable-private-bastion

calvix:
subnet filter tags need to be changed in case of private cluster

tested on goat


mnitchev (approved):

mnitchev:
Looks like that doesn't render though

calvix:
> Looks like that doesn't render though

should be fixed

mnitchev:
Hmm looks like the tests don't trigger either for some reason.
/test create
/test upgrade

PR #185, opened 2022-12-09
remove-containerd-workaround

calvix:
remove workaround fro regsitry container auth
towards https://github.com/giantswarm/roadmap/issues/1737

### DIFF
--- a/helm/cluster-aws/templates/_helpers.tpl
+++ b/helm/cluster-aws/templates/_helpers.tpl
@@ -97,7 +97,7 @@ room for such suffix.
 {{- define "registryFiles" -}}
 {{- if .Values.registry -}}
 {{- if .Values.registry.configure -}}
-- path: /opt/etc/containerd/conf.d/registry-config.toml
+- path: /etc/containerd/conf.d/registry-config.toml
   permissions: "0600"
   contentFrom:
     secret:
--- a/helm/cluster-aws/templates/_helpers.tpl
+++ b/helm/cluster-aws/templates/_helpers.tpl
@@ -95,7 +95,7 @@ room for such suffix.
 - systemctl restart kubelet
 {{- end -}}
 {{- define "registryFiles" -}}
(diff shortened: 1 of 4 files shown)


helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -95,24 +95,16 @@ room for such suffix.
 - systemctl restart kubelet
 {{- end -}}
 {{- define "registryFiles" -}}
-{{- if and .Values.registry .Values.registry.configure -}}
-- path: /opt/registry-config.toml
+{{- if .Values.registry }}
+{{- if .Values.registry.configure -}}
+- path: /opt//etc/containerd/conf.d/registry-config.toml
```
AverageMarcus, line 100:
```suggestion
- path: /opt/etc/containerd/conf.d/registry-config.toml
```

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -95,24 +95,16 @@ room for such suffix.
 - systemctl restart kubelet
 {{- end -}}
 {{- define "registryFiles" -}}
-{{- if and .Values.registry .Values.registry.configure -}}
-- path: /opt/registry-config.toml
+{{- if .Values.registry }}
+{{- if .Values.registry.configure -}}
```
AverageMarcus, line 99: Are we missing an `{{ end }}` ? We've added two if's but no extra end.

AverageMarcus: Looks like just an issue with the PR view. Ignore :)

AverageMarcus (approved):

helm/cluster-aws/templates/_helpers.tpl
calvix: the diff is tricky as the next block si removed but one `{{end}}` is left

calvix:
cluster looks happy

PR #188, opened 2022-12-16
Introduce an empty file to force upgrade of control plane nodes

AverageMarcus:
Dummy change to force a control plane to roll during upgrade. See https://github.com/giantswarm/roadmap/issues/1766#issuecomment-1355063859


marians:
@AverageMarcus Do you want to keep the branch? It's not deleted yet.

AverageMarcus:
Good catch :) Thanks. Deleted.

PR #189, opened 2022-12-20
set AWS NTP server as the default

bdehri:

For private clusters without internet access, it is not possible to use default ntp server. This PR replaces default ntp server with AWS NTP server which is available on local networks. 

### Checklist

### Trigger e2e tests

/test create
/test upgrade


helm/cluster-aws/files/opt/set-aws-ntp.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+AWS_NTP_SERVER="/etc/systemd/timesyncd.conf"
```
fiunchinho, line 3: I don't think this is used.

fiunchinho (approved):

PR #198, opened 2023-01-11
properly-propagate-no-proxy-values

calvix:
towards https://github.com/giantswarm/giantswarm/issues/25259


helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -86,10 +86,10 @@ room for such suffix.
 {{- define "proxyCommand" -}}
 - export HTTP_PROXY={{ $.Values.proxy.http_proxy }}
 - export HTTPS_PROXY={{ $.Values.proxy.https_proxy }}
-- export NO_PROXY=127.0.0.1,localhost,svc,local,169.254.169.254,{{ $.Values.network.vpcCIDR }},{{ $.Values.network.serviceCIDR }},{{ $.Values.network.podCIDR }},{{ include "resource.default.name" $ }}.{{ $.Values.baseDomain }},elb.amazonaws.com
+- export NO_PROXY=127.0.0.1,localhost,svc,local,169.254.169.254,{{ $.Values.network.vpcCIDR }},{{ $.Values.network.serviceCIDR }},{{ $.Values.network.podCIDR }},{{ include "resource.default.name" $ }}.{{ $.Values.baseDomain }},elb.amazonaws.com,{{ $.Values.proxy.no_proxy }}
```
zewolfe, line 89: Just curious about the `$` on this line and its absence on line 92

calvix: good catch, should be both using global context

calvix: fixed in all places

PR #199, opened 2023-01-11
Included an option to add harbor registry to registry-config

DeeAjayi:
Co-authored-by: William Young <will.young@engineerbetter.com>

- Restructured `.Values.registry` to accommodate configuration for docker and Harbor registries this is to enable us to be able to add a private Harbor registry as a mirror for docker in containerd

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,21 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-
 ## Breaking Change
 
-- Restructured `registry.password` and `registry.username` to `registry.dockerRegistry.password` and `registry.dockerRegistry.username`
+- Replaced `registry` parameter  to `connectivity.containerRegistries` in the values schema
 
 ### Added
-- Included harbor registry config to regsitry-config.toml as a mirror for "docker.io" host
-- Add `registry.harborRegistry.url`,  `registry.harborRegistry.password` and `registry.harborRegistry.username` to accomodate harbor regsitry configuration
+
+- Made registry configurations `connectivity.containerRegistries` dynamic to accept as many container registries and mirrors as needed
 
 ## [0.23.0] - 2023-02-01
 
 ### Added
 
 - Add value to specify which AWS account ID to use when associating Route53 Resolver Rules with workload cluster VPC.
-
 
 ## [0.22.0] - 2023-01-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Breaking Change
+
+- Restructured `registry.password` and `registry.username` to `registry.dockerRegistry.password` and `registry.dockerRegistry.username`
+
+### Added
+- Included harbor registry config to regsitry-config.toml as a mirror for "docker.io" host
+- Add `registry.harborRegistry.url`,  `registry.harborRegistry.password` and `registry.harborRegistry.username` to accomodate harbor regsitry configuration
+
 ## [0.22.0] - 2023-01-24
 
 ### Changed
--- a/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
+++ b/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
@@ -14,11 +14,11 @@
       {{ with $value.credentials -}}
       [plugins."io.containerd.grpc.v1.cri".registry.configs."{{$value.endpoint}}".auth]
       {{ if and .username .password -}}
-      auth := {{ printf "%s:%s" .username .password | b64enc | quote }}
+      auth = {{ printf "%s:%s" .username .password | b64enc | quote }}
       {{- else if .auth -}}
       auth = {{ .auth | quote }}
       {{ else if .identitytoken -}}
-      identitytoken := {{ .identitytoken  | quote }}
+      identitytoken = {{ .identitytoken  | quote }}
       {{- end }}
     {{- end }}
   {{- end }}
--- a/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
+++ b/helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
@@ -2,16 +2,16 @@
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
       endpoint = [
-        {{- if and .Values.configure .Values.registry.harborRegistry -}}
+        {{- if and .Values.registry.configure .Values.registry.harborRegistry -}}
         {{ include "fullurl" . | quote }},
         {{- end -}}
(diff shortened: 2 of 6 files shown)


calvix:
Changelog entry is missing

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -1,7 +1,15 @@
 [plugins."io.containerd.grpc.v1.cri".registry]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
+{{- if .Values.harborRegistry -}}
+      endpoint = ["{{ .Values.harborRegistry.username }}"", "https://registry-1.docker.io", "giantswarm.azurecr.io"]
```
AndiDog, line 5: There's a double `"` which makes the config invalid.

Also, do we really need to copy all values in the _else_ branch, or could we put the 2-3 endpoints into multiple lines and only use an `if` clause for harbor?

AndiDog (changes requested):

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -1,7 +1,15 @@
 [plugins."io.containerd.grpc.v1.cri".registry]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-      endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
+      endpoint = [
+        {{- if .Values.harborRegistry -}}
+        "{{ include fullurl }}"
```
MarcelMue, line 6:
```suggestion
        "{{ include fullurl . }}"
```

DeeAjayi: Same response

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -1,7 +1,15 @@
 [plugins."io.containerd.grpc.v1.cri".registry]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-      endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
+      endpoint = [
+        {{- if .Values.harborRegistry -}}
+        {{ include fullurl . | quote }}
```
AverageMarcus, line 6:
```suggestion
        {{ include "fullurl" . | quote }}
```

helm/cluster-aws/values.schema.json
```diff
@@ -254,6 +254,20 @@
                 }
             }
         },
+        "harborRegistry": {
```
MarcelMue, line 257: This should be moved under `registry`

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -1,7 +1,25 @@
...
-    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-      endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
-  [plugins."io.containerd.grpc.v1.cri".registry.configs]
-    [plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]
-      auth = "{{ tpl "{{ .Values.registry.username }}:{{ .Values.registry.password }}" . | b64enc }}"
+  {{- range $host, $config := .Values.connectivity.containerRegistries }}
+    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."{{$host}}"]
+      endpoint = [
+        {{- range $value := $config -}}
+        "https://{{$value.endpoint}}",
```
erkanerol, line 7: Are you sure that the last comma is OK here?

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Breaking Change
+
+- Restructured `registry.password` and `registry.username` to `registry.dockerRegistry.password` and `registry.dockerRegistry.username`
+
+### Added
+- Included harbor registry config to regsitry-config.toml as a mirror for "docker.io" host
+- Add `registry.harborRegistry.url`,  `registry.harborRegistry.password` and `registry.harborRegistry.username` to accomodate harbor regsitry configuration
```
erkanerol, line 16: Is this line up-to-date?

erkanerol:
@MarcelMue As far as I see, this is the same change as we have done for CAPVD. There are a few differences in configuration but it is OK.  Note that there is no harbor specific change in this PR. Harbor config will be in the configuration layer by using this new configuration interface.

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
AndiDog: Since it's TOML, that is fine

DeeAjayi: Yes, I have tested with a private containerd config and it won't be a problem

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
```diff
@@ -1,7 +1,25 @@
 [plugins."io.containerd.grpc.v1.cri".registry]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
-    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-      endpoint = ["https://registry-1.docker.io", "giantswarm.azurecr.io"]
```
AndiDog, line 4: This mirror config for `docker.io` gets removed by this PR, right? Shouldn't we keep it, e.g. by providing a default value? Or is that no longer necessary?

DeeAjayi: It all depends on where we want to put the default values, right now it will have to be configured in the values.yml file

erkanerol: For cloud-director, we don't have any default values in the chart. We configure it with app platform by default for all clusters. 

In terms of inter-provider compatibility, it is OK to set sane defaults for your provider as long as you have the same configuration interface (means values schema).

p.s. We are trying to keep apps/charts giantswarm-agnostic as much as possible. It is not feasible for each case and it is sometimes so costly. Regarding this case specifically, I wouldn't put GS registry here by default but it is up to you.

AndiDog (approved):

helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml
fiunchinho: I think it makes sense keeping the config out of the chart and using our `config` system for that.

MarcelMue:
@giantswarm/team-hydra Are you ok for us to merge this? It has been tested on grizzly.

fiunchinho (approved):

PR #209, opened 2023-01-25
Update: use object for machine pools instead of array

zewolfe:

Towards - https://github.com/giantswarm/roadmap/issues/1614

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Breaking Change
 
-- Use object for `.machinePools` schema instead of array. This is to make it easier to overwrite values when using GitOps.
+- Use object for `.machinePools` schema instead of array. This is to make it easier to overwrite values when using GitOps. For migration steps see the "Upgrading to `v0.23.0`" section in the readme.
 
 ## [0.22.0] - 2023-01-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Use object for `.machinePools` schema instead of array. This is to make it easier to overwrite values when using GitOps.
+
 ## [0.22.0] - 2023-01-24
 
 ### Changed
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -108,74 +108,17 @@
             "type": "object",
             "patternProperties": {
                 "^[a-z0-9]{5,10}$": {
-                    "type": "object",
-                    "properties": {
-                        "availabilityZones": {
-                            "type": "array",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "customNodeLabels": {
-                            "type": "array",
-                            "items": {
-                                "type": "string"
-                            }
-                        },
-                        "customNodeTaints": {
-                            "type": "array",
-                            "items": {
-                                "type": "object",
-                                "properties": {
-                                    "effect": {
-                                        "type": "string",
-                                        "enum": [
-                                            "NoSchedule",
-                                            "PreferNoSchedule",
-                                            "NoExecute"
-                                        ]
-                                    },
-                                    "key": {
-                                        "type": "string"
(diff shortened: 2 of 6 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -107,6 +107,9 @@
         "machinePools": {
             "type": "object"
         },
+        defaultMachinePools: {
```
marians:
```suggestion
        "defaultMachinePools": {
```

helm/cluster-aws/values.schema.json
```diff
@@ -107,6 +107,9 @@
         "machinePools": {
             "type": "object"
         },
```
marians, line 109: I think that `machinePools` should have `patternProperties` defined, as we want to restrict the format of node pools identifiers. See https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties

The pattern should be something like `^[a-z0-9]{5,10}$`.

marians: ... and `patternProperties` is needed in order to specify the properties that any child object here can have. This is a necessity for proper validation.

zewolfe: Thanks  👊🏾

helm/cluster-aws/values.schema.json
zewolfe: I had updated this in a previous commit. I'm guessing maybe a refresh was needed 😃

fiunchinho (approved):
LGTM if tests pass

fiunchinho:
I'm not sure if `/retest` actually works

/test create
/test upgrade

zewolfe:
I'm curious.. would the tests pass if `kubectl-gs` hasn't been updated to template the new changes?

fiunchinho:
> I'm curious.. would the tests pass if `kubectl-gs` hasn't been updated to template the new changes?

Good catch! They won't. You have to update it here https://github.com/giantswarm/test-infra/blob/master/tekton/tasks/create-cluster-capi-pure.yaml#L44

zewolfe:
Thanks! Can I make the change to the task and have the pipeline run that specific version of the task? I feel like making the change to the task now and releasing it would break other test runs in the meantime (whilst this and other PRs are waiting to be merged).

fiunchinho:
> Thanks! Can I make the change to the task and have the pipeline run that specific version of the task? I feel like making the change to the task now and releasing it would break other test runs in the meantime (whilst this and other PRs are waiting to be merged).

I'm afraid that's not possible :cry:

helm/cluster-aws/values.schema.json
```diff
@@ -105,73 +105,78 @@
...
-                        "type": "array",
-                        "items": {
+                        },
+                        "maxSize": {
+                            "type": "integer"
+                        },
+                        "minSize": {
+                            "type": "integer"
+                        },
+                        "name": {
```
AndiDog, line 161: `name` is already the object key. We should forbid that key (see https://stackoverflow.com/a/30725034 for a solution) as it won't be used anymore.

helm/cluster-aws/values.schema.json
```diff
@@ -105,73 +105,78 @@
...
+                            "items": {
+                                "type": "object"
+                            }
                         }
                     }
                 }
             }
         },
+        "defaultMachinePools": {
+            "type": "object"
```
AndiDog, line 178: We should have the same schema here. What about defining it once, and then referencing the schema for both `defaultMachinePools` and `machinePools`. JSON schema has a [`$defs` example](http://json-schema.org/understanding-json-schema/structuring.html#defs).

AndiDog (changes requested):

helm/cluster-aws/values.schema.json
zewolfe: Good shout

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Use object for `.machinePools` schema instead of array. This is to make it easier to overwrite values when using GitOps.
```
fiunchinho, line 12: I was thinking that it may be a good idea to make this breaking change more prominent in the changelog, similar to https://github.com/giantswarm/cluster-aws/blob/master/CHANGELOG.md#breaking-change

AndiDog (approved):

PR #210, opened 2023-01-30
Remove unused `releaseVersion` setting

alex-dabija:

Remove unused `releaseVersion` setting from `values.yaml`. It was set for compatibility reasons to have CAPI and Vintage releases on Vintage MCs, but `v20.0.0-alpha1` is now removed.

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Fail in Helm template if `dnsMode=public` is combined with a `baseDomain` ending with `.internal`
+- Fail in Helm template if `dnsMode=public` is combined with a `baseDomain` ending with `.internal`.
+
+### Removed
+
+- Remove unused `releaseVersion` setting from `values.yaml`.
 
 ## [0.25.1] - 2023-02-16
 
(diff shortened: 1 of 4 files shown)


fiunchinho (approved):

fiunchinho:
Tests worked for me on this PR https://github.com/giantswarm/cluster-aws/pull/211 

so it must be something about the label that breaks cluster creation

AndiDog (approved):

alex-dabija:
> I've been debugging this. The cluster can't be created without the `releaseVersion` label because irsa-operator depends on it.

What does it do with the `releaseVersion` label? Does it have custom logic dependent on the version? Is it used to differentiate between vintage and CAPI clusters?

fiunchinho:
> What does it do with the `releaseVersion` label? Does it have custom logic dependent on the version? Is it used to differentiate between vintage and CAPI clusters?

It's used to determine different things like the right endpoint to use, to calculate the name of the bucket, and so on. There is a `release` object that is passed around.

alex-dabija:
I'm closing this PR because the `releaseVersion` is needed by the IRSA operator.

fiunchinho:
I think this may work @alex-dabija , what do you think? https://github.com/giantswarm/irsa-operator/pull/99

alex-dabija:
> I think this may work @alex-dabija , what do you think? [giantswarm/irsa-operator#99](https://github.com/giantswarm/irsa-operator/pull/99)

Yes, that would work. It's not nice, but anyways the separation in code will be gone.

fiunchinho:
Blocked until we can run automatic tests

PR #223, opened 2023-02-23
add-machine-health-check (#223

calvix:
towards https://github.com/giantswarm/roadmap/issues/1978

probably should be disabled on mc-bootstrap as durign the as they do not recommend to use it during `cluster move`

### DIFF
--- a/helm/cluster-aws/templates/_machine_health_check.tpl
+++ b/helm/cluster-aws/templates/_machine_health_check.tpl
@@ -5,7 +5,7 @@ kind: MachineHealthCheck
 metadata:
   labels:
     {{- include "labels.common" . | nindent 4 }}
-  name: {{ include "resource.default.name" . }}
+  name: {{ include "resource.default.name" . }}-control-plane
   namespace: {{ $.Release.Namespace }}
 spec:
   clusterName: {{ include "resource.default.name" $ }}
@@ -14,6 +14,7 @@ spec:
   selector:
     matchLabels:
       cluster.x-k8s.io/cluster-name: {{ include "resource.default.name" $ }}
+      cluster.x-k8s.io/control-plane: ""
   unhealthyConditions:
   - type: Ready
     status: Unknown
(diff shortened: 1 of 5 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -122,6 +122,15 @@ flatcarAWSAccount: "075585003325"
 # Used to override the AMI lookup and instead use a specific image
 ami: ""
 
+# machine health check settings
+# for control plane nodes only as the check do not support MachinePools yet
+machineHealthCheck:
```
fiunchinho, line 127: I'm not sure about this. I think I'd put `machineHealthCheck` inside `controlPlane`, that way everything related to the control plane is one single place. Similarly, each machinepool could define its own `MachineHealthCheck` and I'd put that inside the `machinePools` value.

helm/cluster-aws/templates/_machine_health_check.tpl
```diff
@@ -0,0 +1,26 @@
...
+  name: {{ include "resource.default.name" . }}-control-plane
+  namespace: {{ $.Release.Namespace }}
+spec:
+  clusterName: {{ include "resource.default.name" $ }}
+  maxUnhealthy: {{ .Values.machineHealthCheck.maxUnhealthy }}
+  nodeStartupTimeout: {{ .Values.machineHealthCheck.nodeStartupTimeout }}
+  selector:
+    matchLabels:
+      cluster.x-k8s.io/cluster-name: {{ include "resource.default.name" $ }}
+      cluster.x-k8s.io/control-plane: ""
```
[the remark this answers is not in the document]
calvix: i think we want different check for CP vs nodes  so having it separate make sense

helm/cluster-aws/values.yaml
calvix: but this raises discussion if we want a customer to be able to change the values for the control plane machine health check as if he configures it wrong it could lead to issues

fiunchinho: Couldn't the customer change it even with the current approach?

calvix: yes , but the issue still stands

fiunchinho (approved):

PR #235, opened 2023-03-09
Move defaults to schema

marians:

Towards https://github.com/giantswarm/roadmap/issues/2139

This PR moves the app default values "source of truth" to `values.schema.json`.

The `values.yaml` file still exists, but it gets generated via `helm-values-gen`.

Caution: Assuming that empty strings are not needed as default values, I skipped those (with the exception of /aws/region, where the missing value caused an error in the chart.)

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Applied normalization using `schemalint normalize`
   - Added property schema for /connectivity/containerRegistries
   - Added property schema for subnetTags objects
+  - Added default values
 
 ## [0.27.0] - 2023-03-01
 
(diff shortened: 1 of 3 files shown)


marians:
Currently failing with:

> `[ERROR] templates/: template: cluster-aws/templates/list.yaml:11:4: executing "cluster-aws/templates/list.yaml" at <include "control-plane" .>: error calling include: template: cluster-aws/templates/_control_plane.tpl:93:72: executing "control-plane" at <.Values.aws.region>: wrong type for value; expected string; got interface {}`

marians:
@fiunchinho FYI: It seems that /aws/region requires an empty string as the default value. Maybe the chart can be improved to check for the existence first? (Context: https://gigantic.slack.com/archives/C04RE4FSTFV/p1678268644076439)

fiunchinho (approved):

PR #236, opened 2023-03-09
Use region defaulting wherever possible

fiunchinho:
### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Use region defaulting wherever possible.
+- Use region defaulting wherever possible, removing `region` from schema.
 
 ## [0.27.0] - 2023-03-01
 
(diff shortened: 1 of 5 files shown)


marians:
If the default value is needed, then we should also remove it from values.schema.json and, as a consequence, from values.yaml.

helm/cluster-aws/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: cluster-aws
 description: A helm chart for creating Cluster API clusters with the AWS infrastructure provider (CAPA).
 icon: https://s.giantswarm.io/app-icons/aws/1/dark.svg
 type: application
-version: [[ .Version ]]
+version: 0.1.0
```
AndiDog, line 6: This change shouldn't be committed

fiunchinho: Ooops! Good catch.

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Added property schema for subnetTags objects
   - Added default values
 
+### Fixed
+
+- Use region defaulting wherever possible.
```
marians, line 21:
```suggestion
- Use region defaulting wherever possible, removing `region` from schema.
```

marians (approved):

PR #239, opened 2023-03-10
Add workflow for schema validation

marians:

This PR adds preliminary schema checks to this repository's GitHub actions, to ensure schema validity and avoid round trips. As we have a lot of movement in the schema currently, this is needed.

Note: the plan is to replace this simplistic version soon by one that is maintained automatically via devctl.

### Checklist

### Trigger e2e tests

/test create
/test upgrade


marians:
Currently fails because the schema is not completely normalized. Will be fixed via https://github.com/giantswarm/cluster-aws/pull/238

.github/workflows/json_schema.yaml
```diff
@@ -0,0 +1,43 @@
...
+          go-version: "^1.19"
+
+      - name: Install schemalint
+        run: |
+          go install github.com/giantswarm/schemalint@latest
+      - name: Run schemalint
+        run: |
+          schemalint verify ./helm/cluster-aws/values.schema.json --rule-set cluster-app
+
+  generate:
```
fiunchinho, line 25: `generate` sounds like we are generating something on this action, but we are really validating that the values were generated from the schema. I'd use a different name, but I'm bad at naming. maybe `generate-diff`? I don't know.

marians: True. I'm proposing `values-diff`.

fiunchinho (approved):

marians:
@fiunchinho Can we merge this without e2e tests? (I think so)

fiunchinho:
> @fiunchinho Can we merge this without e2e tests? (I think so)

yes

PR #240, opened 2023-03-10
Create providerSpecific root object and move properties

marians:

Here we populate the `/providerSpecific` object with properties that are AWS specific.

### Checklist

### Trigger e2e tests

/test create
/test upgrade

### DIFF
--- a/helm/cluster-aws/values.yaml
+++ b/helm/cluster-aws/values.yaml
@@ -1,6 +1,4 @@
-aws:
-  awsClusterRoleIdentityName: default
-  region: ""
+aws: {}
 bastion:
   enabled: true
   instanceType: t3.small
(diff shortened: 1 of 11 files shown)


helm/cluster-aws/values.yaml
```diff
@@ -1,5 +1,4 @@
-aws:
-  awsClusterRoleIdentityName: default
+aws: {}
```
fiunchinho, line 1: is this really required?

marians: I'll check

marians: So right now this is still generated by `helm-values-gen`. Regardless whether it's required or not, I would like to keep it for now and get rid of this and other empty objects one the tool is updated.

fiunchinho (approved):

PR #250, opened 2023-03-17
Create /internal object with initial set of properties

marians:

This PR moves first properties into the /internal object, which is dedicated to settings that are meant for Giant Swarm staff only, or which should not exposed via user interfaces, or which are considered unstable/experimental.

### Checklist


zewolfe (approved):
LGTM

PR #252, opened 2023-03-20
Forbid additional properties on the root level

marians:

Towards https://github.com/giantswarm/roadmap/issues/2168

Cluster app values schemas must not allow arbitrary values on the root level, so that our user interface presents a clean overview of all properties in their categories.

This PR disallows arbitrary root level propoerties and makes the required schema changes to keep the chart valid.

### Checklist

### DIFF
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -91,6 +91,7 @@
         }
     },
     "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "additionalProperties": false,
     "properties": {
         "baseDomain": {
             "title": "Base DNS domain",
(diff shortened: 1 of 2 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -91,11 +91,16 @@
         }
     },
     "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "additionalProperties": false,
     "properties": {
         "baseDomain": {
             "title": "Base DNS domain",
             "type": "string"
         },
+        "cluster-shared": {
```
AndiDog, line 100: Where is this used?

marians: Since the cluster-shared app is used as a sub-chart, helm injects `.Values.cluster-shared` here. If we don't have it defined in the schema (because of additionalProperties: false) the chart would fail now.

fiunchinho (approved):

PR #259, opened 2023-03-20
Change nodepools schema to array

marians:

This moves the node pools definition from `/machineDeployments` to `/nodePools`.

As we currently cannot support objects with `patternProperties` in the UI, we also change towards array type.

Additional changes:

- Remove defaultMachinePools, since the default value for /nodePools already includes this
- Set default values for some properties, so that they are exposed in the UI
- Set constraints for some properties, so that some errors are prevented

### Checklist


marians:
Not pursuing this currently. We are looking for different solutions.

PR #264, opened 2023-03-22
Add constraints to properties

marians:

To avoid simpe input mistakes, this PR adds constraints to numeric and string properties.

### Checklist

### DIFF
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -678,7 +678,7 @@
                     "description": "If specified, this image will be used to provision EC2 instances.",
                     "maxLength": 100,
                     "minLength": 8,
-                    "pattern": "^ami-[0-9abzdefg]+$",
+                    "pattern": "^ami-[0-9abcdefg]+$",
                     "title": "Amazon machine image (AMI)",
                     "type": "string"
                 },
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -243,19 +243,22 @@
                             "default": "100.64.0.0/12",
                             "description": "IPv4 address range for pods, in CIDR notation.",
                             "title": "Pod subnet",
-                            "type": "string"
+                            "type": "string",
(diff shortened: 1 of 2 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -630,10 +643,15 @@
                 },
                 "name": {
                     "description": "Unique identifier, cannot be changed after creation.",
+                    "maxLength": 10,
```
AndiDog, line 646: I also saw these numbers for vintage clusters. Do we _really_ need to restrict at 10 characters? If not, I'd rather leave this at Kubernetes' maximum name limit. Same for the minimum.

helm/cluster-aws/values.schema.json
```diff
@@ -658,6 +676,9 @@
             "properties": {
                 "ami": {
                     "description": "If specified, this image will be used to provision EC2 instances.",
+                    "maxLength": 100,
+                    "minLength": 8,
+                    "pattern": "^ami-[0-9abzdefg]+$",
```
AndiDog, line 681: Where's this regex from? Do they only use this alphabet choice?

marians: I looked at documentation and all examples I found used a hexadecimal ID with the `ami-` prefix prepended.

fiunchinho:
What's the state of this PR @marians ? I see that it's still a draft.

helm/cluster-aws/values.schema.json
marians, line 681:
```suggestion
                    "pattern": "^ami-[0-9abcdefg]+$",
```

marians:
I have parked this PR since I didn't find the time to work out which constraints would actually make sense. But I think it's still needed, as we have many strings where e. g. adding a blank space by mistake could break the whole thing and this would be easy to catch.

AndiDog:
Let's close this since the diff is outdated. We'd have to redo the change once time allows and if we still find it useful.

PR #267, opened 2023-03-24
Add migration help to changelog

marians:

This PR adds help to migrate values from v0.27.0 to v0.28.0.

As a prerequisite, the user needs the pure values as a local YAML file, and `yq` installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,68 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.28.0] - 2023-03-23
 
+**Note**: this release includes values schema changes which break compatibility with previous versions.
+
+<details>
+<summary>How to migrate from v0.27.0</summary>
+  
+To migrate values from cluster-aws v0.27.0, we provide below [yq](https://mikefarah.gitbook.io/yq/) script, which assumes your values (not a ConfigMap!) are available in the file `values.yaml`. Note that the file will be overwritten.
+  
+Also be aware that if you were using `.aws.awsClusterRole` to specify a role in v0.27.0, this cannot be migrated automatically. Instead you have to make sure to have a [AWSClusterRoleIdentity](https://cluster-api-aws.sigs.k8s.io/topics/multitenancy.html#awsclusterroleidentity) resource in the management cluster which specifies the identity to use. The name of that resource then has to be specified as `.providerSpecific.awsClusterRoleIdentityName` in the new values for v.28.0.
+  
+```bash
+yq --inplace '
+  .connectivity.availabilityZoneUsageLimit = .network.availabilityZoneUsageLimit |
+  .connectivity.bastion = .bastion |
+  .connectivity.dns.additionalVpc = (.network.dnsAssignAdditionalVPCs | split(",")) |
+  .connectivity.dns.mode = .network.dnsMode |
+  .connectivity.dns.resolverRulesOwnerAccount = .network.resolverRulesOwnerAccount |
+  .connectivity.network.podCidr = .network.podCIDR |
+  .connectivity.network.serviceCidr = .network.serviceCIDR |
+  .connectivity.network.vpcCidr = .network.vpcCIDR |
+  .connectivity.proxy.httpProxy = .proxy.http_proxy |
+  .connectivity.proxy.httpsProxy = .proxy.https_proxy |
+  .connectivity.proxy.noProxy = .proxy.no_proxy |
+  .connectivity.sshSsoPublicKey = .sshSSOPublicKey |
+  .connectivity.subnets = .network.subnets |
+  .connectivity.topology.mode = .network.topologyMode |
+  .connectivity.topology.prefixListId = .network.prefixListID |
+  .connectivity.topology.transitGatewayId = .network.transitGatewayID |
+  .connectivity.vpcEndpointMode = .network.vpcEndpointMode |
+  .connectivity.vpcMode = .network.vpcMode |
+  .controlPlane.apiMode = .network.apiMode |
+  .controlPlane.oidc = .oidc |
+  .internal.hashSalt = .hashSalt |
+  .internal.kubernetesVersion = .kubernetesVersion |
+  .metadata.description = .clusterDescription |
+  .metadata.name = .clusterName |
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,68 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.28.0] - 2023-03-23
 
+**Note**: this release includes values schema changes which break compatibility with previous versions.
```
AverageMarcus, line 12: It might be better to include this in the project README.md and link to it from here. There's a couple examples of this already e.g. https://github.com/giantswarm/cluster-aws/blob/migration-help/CHANGELOG.md#breaking-change-1

marians: Honestly I think the changelog is the better place for such migration info. Especially once we have values documentation in the README, which will make the README quite long. I'm sure only few people will expect this down there. So I'd rather vote for moving all migration info to the changelog, which could be done in a separate PR.

AndiDog: I agree that a strong hint about a breaking change in the _changelog_ is where people would look at, including our own developers.

For large-scale changes like these, we have to ensure that people cannot accidentally use the old values. Since we now have `required "connectivity.dns.mode required" .Values.connectivity.dns.mode`, an old values.yaml becomes broken and leads to a loud error as it should be, instead of creating a cluster with wrong values 👍.

Can we replace `## Changed` with `## Breaking changes` as well?

AndiDog (commented):
LGTM so far

AndiDog (changes requested):
Found one problem with the migration script – unset values are explicitly set to `null`. In my case, I got this as output:

```yaml
controlPlane:
  replicas: 3
  apiMode: private
  oidc: null
connectivity:
  availabilityZoneUsageLimit: 3
  bastion: {}
  dns:
    additionalVpc: null
    mode: private
    resolverRulesOwnerAccount: null
  network:
    podCidr: null
    serviceCidr: null
    vpcCidr: 10.226.0.0/18
  proxy:
    httpProxy: http://[...]
    httpsProxy: http://[...]
    noProxy: null
  sshSsoPublicKey: null
  subnets:
    - cidrBlocks:
        - availabilityZone: a
          cidr: 10.226.0.0/20
        - availabilityZone: b
          cidr: 10.226.16.0/20
        - availabilityZone: c
          cidr: 10.226.32.0/20
      isPublic: false
  topology:
    mode: GiantSwarmManaged
    prefixListId: null
    transitGatewayId: null
  vpcEndpointMode: null
  vpcMode: private
internal:
  hashSalt: null
  kubernetesVersion: null
metadata:
  description: Andreas Sommer's test cluster
  name: andreas42
  organization: giantswarm
nodePools:
  machine-pool0:
    instanceType: m5.xlarge
    maxSize: 10
    minSize: 3
    rootVolumeSizeGB: 300
providerSpecific:
  ami: null
  awsClusterRoleIdentityName: null
  flatcarAwsAccount: null
  region: null
```

CHANGELOG.md
```diff
@@ -13,6 +13,68 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+<details>
+<summary>How to migrate from v0.27.0</summary>
+  
+To migrate values from cluster-aws v0.27.0, we provide below [yq](https://mikefarah.gitbook.io/yq/) script, which assumes your values (not a ConfigMap!) are available in the file `values.yaml`. Note that the file will be overwritten.
+  
+Also be aware that if you were using `.aws.awsClusterRole` to specify a role in v0.27.0, this cannot be migrated automatically. Instead you have to make sure to have a [AWSClusterRoleIdentity](https://cluster-api-aws.sigs.k8s.io/topics/multitenancy.html#awsclusterroleidentity) resource in the management cluster which specifies the identity to use. The name of that resource then has to be specified as `.providerSpecific.awsClusterRoleIdentityName` in the new values for v.28.0.
```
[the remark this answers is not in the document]
marians: This change was not only about the property name. The meaning of the property has also changed. With `.aws.awsClusterRole` the name of the role had to be set, while `.providerSpecific.awsClusterRoleIdentityName` is the name of a kubernetes resource. By adding "the deprecated name" this would become less clear, I think.

CHANGELOG.md
marians: Not all values schema changes were actually breaking. With the note above, is it clear enough that there are incompatible changes?

calvix (approved):
looks good

AndiDog:
> Approval if `additionalVpc` is on purpose 🤷

Not sure how to handle this: our code uses singular by mistake despite it being a comma-separated list in our implementation (annotation `const AWSDNSAdditionalVPC = "aws.giantswarm.io/dns-assign-additional-vpc"`, dns-operator-aws allows multiple comma-separated values in there).

CHANGELOG.md
```diff
@@ -15,6 +15,68 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  with(select(.bastion != null);                            .connectivity.bastion = .bastion) |
+  with(select(.clusterDescription != null);                 .metadata.description = .clusterDescription) |
+  with(select(.clusterName != null);                        .metadata.name = .clusterName) |
+  with(select(.flatcarAWSAccount != null);                  .providerSpecific.flatcarAwsAccount = .flatcarAWSAccount) |
+  with(select(.hashSalt != null);                           .internal.hashSalt = .hashSalt) |
+  with(select(.kubernetesVersion != null);                  .internal.kubernetesVersion = .kubernetesVersion) |
+  with(select(.machinePools != null);                       .nodePools = .machinePools) |
+  with(select(.network.apiMode != null);                    .controlPlane.apiMode = .network.apiMode) |
+  with(select(.network.availabilityZoneUsageLimit != null); .connectivity.availabilityZoneUsageLimit = .network.availabilityZoneUsageLimit) |
+  with(select(.network.dnsAssignAdditionalVPCs != null);    .connectivity.dns.additionalVpc = (.network.dnsAssignAdditionalVPCs | split(","))) |
```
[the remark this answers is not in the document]
marians: This is the actual name of the property. Since `VPC` is an acronym, it's not necessary to add an `s` for a plural form.

PR #271, opened 2023-03-31
enable-flatcar-os-bastion

calvix:
use CAPI bootstrap controller to manage flatcar for bastion node
towards https://github.com/giantswarm/roadmap/issues/996


helm/cluster-aws/files/etc/ssh/sshd_config
```diff
@@ -13,3 +13,4 @@ MaxAuthTries 5
 LoginGraceTime 60
 AllowTcpForwarding no
 AllowAgentForwarding no
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
```
calvix, line 16: still necessary as the flatcar is denying RSA by default

helm/cluster-aws/templates/_bastion.tpl
```diff
@@ -99,4 +87,31 @@ metadata:
   name: {{ include "resource.default.name" $ }}-bastion-{{ include "hash" (dict "data" (include "bastion-awsmachinetemplate-spec" $) "global" .) }}
   namespace: {{ .Release.Namespace }}
 spec: {{ include "bastion-awsmachinetemplate-spec" $ | nindent 2 }}
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
```
fiunchinho, line 92: Has this object a similar immutability to other CAPA `*Template` CRs like `AWSMachineTemplate`? For those we create a different CR when changing it. Is that necessary here?

calvix: Well we do not have it for  `KubeadmConfig` for the MachinePool but I found something similar for the MachineDeployment in ` cluster-gcp` so maybe its worth integrating similar approach

fiunchinho: I would only do it if there are immutable fields, otherwise it's just adding more complexity.

calvix: I think it is the case as we needed it for the MD in gcp

fiunchinho (approved):

PR #273, opened 2023-04-03
Fix rendering `oidc.pem` by mistake when not specified

AndiDog:

https://github.com/giantswarm/giantswarm/issues/26480

The file gets rendered by mistake if the value is `nil` instead of `""`.

Not sure what implications this has, but we already saw some hard errors like `missing content for CA bundle`. Possibly cluster creation is also affected.

PR not tested yet with cluster creation at time of opening it. Will do that now.

### Checklist

### Trigger e2e tests

/test create
/test upgrade


AverageMarcus (approved):
LGTM

PR #276, opened 2023-04-05
Configure kubelet Shutdown Grace Period 

mnitchev:

This PR configures both the `shutdownGracePeriod` and `shutdownGracePeriodCriticalPods`. The grace period is set to 5 minutes while the critical pods grace period is set to 1 minute. This means that kubelet will start terminating critical pods (pods with Priority class set to `system-cluster-critical` or `system-node-critical`) in the last 1 minute of the `shutdownGracePeriod`. This is important since our CNI (cilium) is a critical pod that needs to terminated last.

This feature is implemented using [systemd Inhibitor Locs](https://www.freedesktop.org/wiki/Software/systemd/inhibit). The maximum inhibit delay is controlled by the `InhibitDelayMaxSec` in `logind`. On our AWS nodes the default is set to 30 seconds, so we need to override it. This is why the `/lib/systemd/logind.conf.d/zzz-kubelet-graceful-shutdown.conf` is mounted in the nodes (the `zzz` prefix is so it always lands as the last evaluated file).

### Checklist

### References
1. [Graceful Node Shutdown docs](https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/)
2. [Graceful Node Shutdown Proposal docs](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2000-graceful-node-shutdown/README.md) (there's some more info on how it works in here)
3. [logind.conf docs](https://www.freedesktop.org/software/systemd/man/logind.conf.html)
4. [systemd inhibit docs](https://www.freedesktop.org/wiki/Software/systemd/inhibit/)

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- Configure kubelet `ShutdownGracePeriod` to 5m and `ShutdownGracePeriodCriticalPods` to 1m.
+- Set default Node systemd logind `InhibitDelayMaxSec` to 5m.
+
 ## [0.29.0] - 2023-03-27
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- Configure kubelet `ShutdownGracePeriod` to 5m and `ShutdownGracePeriodCriticalPods` to 1m.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- Configure kubelet `ShutdownGracePeriod` to 5m and `ShutdownGracePeriodCriticalPods` to 1m.
```
AndiDog, line 12: Let's add a high-level description here so customers understand the rationale

AndiDog (commented):
LGTM apart from comments

CHANGELOG.md
mnitchev: I added a bit more of an explanation. Please take a look

AndiDog (approved):

PR #281, opened 2023-04-12
Default to using `giantswarm.azurecr.io` as Docker Hub mirror

AndiDog:

Towards https://github.com/giantswarm/roadmap/issues/1912

Add mirror endpoint for `docker.io/...` images by default. With this, Docker Hub rate limit errors get hidden (not solved).

I've put the default directly into `connectivity.containerRegistries` instead of introducing a field `internal.defaultContainerRegistries`. That's because we want to _merge_ if customers provide their own. For example, `connectivity: { containerRegistries: { my-harbor.example.com: { ... } } }` is imaginable and should leave our defaults for `docker.io` in place. However, if the key `docker.io` is customized, the value gets overwritten since it's an array.

### Checklist

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Rename `defaultMachinePools` to `internal.nodePools` to fit new schema requirements and make clear that it should not be changed by customers
+- Default to using `giantswarm.azurecr.io` as Docker Hub mirror
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


fiunchinho (changes requested):
[This is the mirror configuration](https://github.com/giantswarm/config/blob/main/default/config.yaml#L7) created by honeybadger, the team owning the registries. If I understand correctly, we want to first pull from Docker, then from azure. But we don't want to pull from quay.

My information my be outdated at this point, but I thought we wanted to stop using quay.

PR #288, opened 2023-04-25
Remove old JSON schema workflow

mogottsch:

The JSON schema workflow is managed through devctl and we don't need the old one anymore.

### Checklist


fiunchinho (approved):
Thanks!!

PR #294, opened 2023-04-27
Add coredns, cni and csi as helmreleases

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/2414

### Checklist

### Trigger e2e tests

Not yet.


calvix (approved):

calvix:
how do we handle providing a proxy values to a cloud manager to the helm release?

fiunchinho:
> how do we handle providing a proxy values to a cloud manager to the helm release?

It's not handled at the moment. I'm thinking about using [the values that we already define here](https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.schema.json#L271-L272).

calvix:
what is blocking this PR? is breaking our mc bootstrap CI so i wonder if we need some workaround fix for now until we can merge this

calvix:
> It's blocked by this https://github.com/giantswarm/giantswarm/issues/26735. Why is an unmerged cluster-aws PR breaking MC bootstrap ?

you introduced `aws-cloud-controller-manager` as an app into `default-apps-aws` which means the nodes are marked as ready only after this app is installed via `default-apps-aws` which is only at the end of the setup when we deploy management cluster fleet.   And in mc bootstrap we want to have  reliable api before we start running helm againts it so we wait until all 3 control plane nodes are ready, which will never happen unless the `aws-cloud-controller-manager` is installed


moving it on `cluster-aws` will fix the problem as it will be installed as part of the cluster bootstrap

fiunchinho:
So this PR is a potential fix to the problem, it's not the one breaking mc-bootstrap as you mentioned. That makes more sense.

calvix (approved):

PR #298, opened 2023-05-01
Use fixed alias cloudfront domain for IRSA

tuladhar:
Towards - https://github.com/giantswarm/roadmap/issues/2420

### Checklist

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites


calvix:
why use that complex solution? why not simply replace the value in the KubeadmControlPlane rather then sue files and replacer?

AndiDog:
This depends on a change in `irsa-operator` to unconditionally create an alias domain. Is there a PR for that?

tuladhar:
> This depends on a change in `irsa-operator` to unconditionally create an alias domain. Is there a PR for that?

Not yet, Andreas. I discussed with Nick, and currently alias domain is only created after CloudFront + ACM certificate is fully 
provisioned. 

⚠️  We don't know and need to test how control plane will behave when we set `service-account-issuer` URL to `irsa` alias domain which doesn't exist yet (NX domain)

AndiDog:
True. What I meant is that irsa-operator would only create a domain for public clusters:

```
func (s *Service) getCloudFrontAliasDomain(ctx context.Context, baseDomain string) (string, error) {
	awscluster := &capa.AWSCluster{}
	err := s.Client.Get(ctx, types.NamespacedName{Namespace: s.Scope.ClusterNamespace(), Name: s.Scope.ClusterName()}, awscluster)
	if apierrors.IsNotFound(err) {
		// fallthrough
	} else if err != nil {
		return "", err
	} else {
		private := awscluster.Annotations["aws.giantswarm.io/vpc-mode"]
		if private != "private" {
			return key.CloudFrontAlias(baseDomain), nil
		}
	}
	return "", nil
}
```

bdehri:
@njuettner Could you please take a quick look?

njuettner (approved):
lgtmiiwfy

PR #301, opened 2023-05-03
Move podCidr and serviceCidr to new properties

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/26798

### Checklist

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites

### DIFF
--- a/helm/cluster-aws/templates/_cluster.tpl
+++ b/helm/cluster-aws/templates/_cluster.tpl
@@ -22,11 +22,9 @@ metadata:
 spec:
   clusterNetwork:
     services:
-      cidrBlocks:
-       - {{ .Values.connectivity.network.serviceCidr }}
+      cidrBlocks: {{ .Values.connectivity.clusterNetwork.services.cidrBlocks }}
     pods:
-      cidrBlocks:
-      - {{ .Values.connectivity.network.podCidr }}
+      cidrBlocks: {{ .Values.connectivity.clusterNetwork.pods.cidrBlocks }}
   controlPlaneRef:
     apiVersion: controlplane.cluster.x-k8s.io/v1beta1
     kind: KubeadmControlPlane
--- a/helm/cluster-aws/templates/_cluster.tpl
+++ b/helm/cluster-aws/templates/_cluster.tpl
@@ -22,9 +22,11 @@ metadata:
 spec:
   clusterNetwork:
     services:
-      cidrBlocks: {{ .Values.connectivity.network.services.cidrBlocks }}
+      cidrBlocks:
+      {{- toYaml .Values.connectivity.network.services.cidrBlocks | nindent 8 }}
     pods:
-      cidrBlocks: {{ .Values.connectivity.network.pods.cidrBlocks }}
+      cidrBlocks:
+      {{- toYaml .Values.connectivity.network.pods.cidrBlocks | nindent 8 }}
   controlPlaneRef:
     apiVersion: controlplane.cluster.x-k8s.io/v1beta1
     kind: KubeadmControlPlane
--- a/helm/cluster-aws/templates/_cluster.tpl
+++ b/helm/cluster-aws/templates/_cluster.tpl
@@ -22,9 +22,9 @@ metadata:
 spec:
   clusterNetwork:
     services:
-      cidrBlocks: {{ .Values.connectivity.clusterNetwork.services.cidrBlocks }}
+      cidrBlocks: {{ .Values.connectivity.network.services.cidrBlocks }}
     pods:
-      cidrBlocks: {{ .Values.connectivity.clusterNetwork.pods.cidrBlocks }}
+      cidrBlocks: {{ .Values.connectivity.network.pods.cidrBlocks }}
   controlPlaneRef:
     apiVersion: controlplane.cluster.x-k8s.io/v1beta1
     kind: KubeadmControlPlane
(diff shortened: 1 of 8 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -145,6 +145,68 @@
                         }
                     }
                 },
+                "clusterNetwork": {
+                    "type": "object",
+                    "title": "ClusterNetwork",
```
marians, line 150: Titles should always be sentence case.

```suggestion
                    "title": "Cluster network",
```

marians (changes requested):
Having two network sections on the same level will be extremely confusing. Let's please fix that!

image

fiunchinho:
This PR is blocked until we reach agreement here https://github.com/giantswarm/giantswarm/issues/26451

helm/cluster-aws/values.yaml
```diff
@@ -13,8 +13,12 @@ connectivity:
   dns:
     mode: public
   network:
-    podCidr: 100.64.0.0/12
-    serviceCidr: 172.31.0.0/16
```
nprokopic, line 17: Not sure how important (depending on the number of existing clusters out there and who and when will be upgrading them), maybe not important at all, removing these could break the existing clusters if they do not immediately set the new values when upgrading the version of `cluster-aws`.

Something like this could be smoother and more friendly to backward compatibility (if that's even needed):
- Change 1:
  - Add new values fields
  - In templates use new fields and fallback to old fields if new ones are not specified
  - Deprecate old values fields
- Leave both old and new fields for few releases
- Change 2:
  - Remove old values fields
  - Use just the new fields

fiunchinho: We can apply the changes at the same time when creating the PR on `management-clusters-fleet`. That's how we've done it in the past. I think that's fine while we manage the clusters. Once we don't manage them and the customers start managing them, then we need to add backwards compatible changes. But for now I think it's fine. Does that make sense?

nprokopic: Yup, makes sense :)

marians:
FYI: the current way looks good to me :)

image

marians:
FYI: the current way looks good to me :)

image

fiunchinho:
hey @nprokopic @marians any reasons not to approve the PR? is there anything missing?

nprokopic (approved):

nprokopic:
> hey @nprokopic @marians any reasons not to approve the PR? is there anything missing?

Forgot to click the button yesterday :grimacing:

marians (approved):
The change looks like an improvement to me.

Since this change is breaking schema compatibility, I recommend to provide a script/command like we did in v0.28.0 (see CHANGELOG.md).

fiunchinho:
> The change looks like an improvement to me.
> 
> Since this change is breaking schema compatibility, I recommend to provide a script/command like we did in v0.28.0 (see CHANGELOG.md).

I think that's a good idea generally. But in this case I don't think it's necessary https://github.com/giantswarm/cluster-aws/pull/301#discussion_r1189719020

helm/cluster-aws/values.schema.json
```diff
@@ -251,17 +251,57 @@
...
                     "properties": {
-                        "podCidr": {
-                            "type": "string",
-                            "title": "Pod subnet",
-                            "description": "IPv4 address range for pods, in CIDR notation.",
-                            "default": "100.64.0.0/12"
+                        "pods": {
+                            "type": "object",
+                            "title": "Pods",
+                            "required": [
```
AndiDog, line 257: [repeated]

Is it `required` if it has a `default`? (see [R9 in the schema RFC](https://github.com/giantswarm/rfc/pull/55/files))

helm/cluster-aws/templates/_control_plane.tpl
```diff
@@ -142,7 +142,7 @@ spec:
             listen-metrics-urls: "http://0.0.0.0:2381"
             quota-backend-bytes: "8589934592"
       networking:
-        serviceSubnet: {{ .Values.connectivity.network.serviceCidr }}
+        serviceSubnet: {{ join "," .Values.connectivity.network.services.cidrBlocks }}
```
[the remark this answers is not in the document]
fiunchinho: [According to the docs](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api/bootstrap.cluster.x-k8s.io/KubeadmConfig/v1beta1@v1.4.1#spec-clusterConfiguration-networking-serviceSubnet) it accepts a comma-delimited string.

helm/cluster-aws/values.schema.json
fiunchinho: Fixed

helm/cluster-aws/templates/_cluster.tpl
```diff
@@ -22,11 +22,9 @@ metadata:
 spec:
   clusterNetwork:
     services:
-      cidrBlocks:
-       - {{ .Values.connectivity.network.serviceCidr }}
+      cidrBlocks: {{ .Values.connectivity.network.services.cidrBlocks }}
```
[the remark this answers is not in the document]
fiunchinho: Fixed.

kopiczko (approved):

marians:
@fiunchinho I think the repo README.md includes some examples that should be adapted to this schema change.

AndiDog (approved):

fiunchinho:
> @fiunchinho I think the repo README.md includes some examples that should be adapted to this schema change.

I don't think we mention the changed fields in the examples.

PR #310, opened 2023-05-24
Use m6.2xlarge as default instance type

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/1863

### Checklist

### Trigger e2e tests

/test upgrade
/run cluster-test-suites


whites11 (approved):

njuettner:
In case we don't need so much CPU but Memory we could also go for the latest r6 instance type

```
r6i.xlarge | 4 | 32 GiB | Up to 12500 Megabit | EBS only | 0.304
r6i.2xlarge |	8 |	64 GiB	| Up to 12500 Megabit |	EBS only |	0.608
```

PR #322, opened 2023-06-14
Flatcar nodes

calvix:
* CP and worker nodes machines on flatcar
* move several scripts into the proper configuration 
*  adjust  OS setting and hardening to match what we have in k8scloudconfig

It is partially breaking change as `KubeadmControlPlane` webhook does not allow change of the format from `cloud-init` to `ignition`, it can be worked around by shortly disabling the webhook and letting chart operator do the update of CRs, also the worker nodes became unavailable for a while as it takes a while until the ignition is rendered and in that time the operator tries to launch Flatcar OS and feeding it the cloud-init instead of ignition. But that will resolve after 5-10 minutes and workers will join the cluster


helm/cluster-aws/files/etc/containerd/config.toml
```diff
@@ -1,3 +1,27 @@
...
+no_shim = false
+
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+# setting runc.options unsets parent settings
+runtime_type = "io.containerd.runc.v2"
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+SystemdCgroup = true
+[plugins."io.containerd.grpc.v1.cri"]
+sandbox_image = "{{ .Values.internal.sandboxContainerImage.registry }}/{{ .Values.internal.sandboxContainerImage.name }}:{{ .Values.internal.sandboxContainerImage.tag }}"
+
```
calvix, line 24: to avoid a weird bug where containerd kills random pods - https://github.com/giantswarm/roadmap/issues/1737
also configure sandbox pause container to use our registry instead of k8s gcr

helm/cluster-aws/files/etc/systemd/timesyncd.conf
```diff
@@ -0,0 +1,2 @@
+[Time]
+NTP=169.254.169.123
```
calvix, line 2: configure timesyncd instead of the ntp script

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -148,19 +139,79 @@ room for such suffix.
...
+    device: /dev/xvdd
+    wipeFilesystem: true
+    label: containerd
+    format: xfs
+- name: kubelet
+  mount:
+    device: /dev/xvde
+    wipeFilesystem: true
+    label: kubelet
+    format: xfs
```
calvix, line 171: configure disk with ignition instead of bash script

helm/cluster-aws/files/etc/containerd/config.toml
```diff
@@ -1,3 +1,27 @@
+version = 2
+
+# set containerd as a subreaper on linux when it is not running as PID 1
+subreaper = true
+# set containerd's OOM score
```
AndiDog, line 5: Can we add a comment that explains the value?

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -99,14 +93,11 @@ room for such suffix.
 - export http_proxy={{ $.Values.connectivity.proxy.httpProxy }}
 - export https_proxy={{ $.Values.connectivity.proxy.httpsProxy }}
 - export no_proxy=127.0.0.1,localhost,svc,local,169.254.169.254,{{ $.Values.connectivity.network.vpcCidr }},{{ join "," $.Values.connectivity.network.services.cidrBlocks }},{{ join "," $.Values.connectivity.network.pods.cidrBlocks }},{{ include "resource.default.name" $ }}.{{ $.Values.baseDomain }},elb.amazonaws.com,{{ $.Values.connectivity.proxy.noProxy }}
-- systemctl daemon-reload
-- systemctl restart containerd
-- systemctl restart kubelet
 {{- end -}}
 
 {{- define "registryFiles" -}}
-- path: /etc/containerd/conf.d/registry-config.toml
```
AndiDog, line 108: Delete `helm/cluster-aws/files/etc/containerd/conf.d/registry-config.toml`?!

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -187,17 +234,28 @@ ami:
   id: {{ . | quote }}
 {{- else -}}
 ami: {}
-imageLookupBaseOS: "ubuntu-20.04"
+imageLookupBaseOS: "flatcar-stable"
```
AndiDog, line 237: Can we / should we pin to a specific version?

helm/cluster-aws/files/etc/sysctl.d/hardening.conf
```diff
@@ -0,0 +1,26 @@
+fs.inotify.max_user_watches = 16384
+# Default is 128, doubling for nodes with many pods
+# See https://github.com/giantswarm/giantswarm/issues/7711
+fs.inotify.max_user_instances = 8192
```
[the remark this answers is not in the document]
calvix: that comes from the conf we had in vintage, I didn't go thru all the values but as the comment is 4 years old I assume we had to increase it a few more times, I will remove the misleading comment but leave the value

helm/cluster-aws/templates/_helpers.tpl
calvix: It's deleted but github shows it as rename from original file to  `helm/cluster-aws/files/etc/containerd/config.toml`

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -187,17 +234,28 @@ ami:
...
-{{- define "prepare-varLibKubelet-Dir" -}}
-- /bin/test ! -d /var/lib/kubelet && (/bin/mkdir -p /var/lib/kubelet && /bin/chmod 0750 /var/lib/kubelet)
+{{- define "nodeDirectories" -}}
+- path: /var/lib/kubelet
+  mode: 0750
 {{- end -}}
 
-{{- define "flatcarKubeadmService" -}}
+{{- define "flatcarSystemdUnits" -}}
+- name: kubereserved.slice
```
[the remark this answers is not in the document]
calvix: https://github.com/giantswarm/cluster-aws/pull/322/files#diff-72301810e9b39353e074da2c29eadb7e61d61ca28007476794b2a07103336500R281

its a slice used for containers running with containerd so they are separate from the global slice its basically separate cgroup for the containers, again its something useful we did a long time ago in vintage but it wasn't easily applicable with the cloud-init in ubuntu

for example if you run `systemd-cgtop` you can easily see what resources the  kubereserved slice eat meaning what resources  use containers compared to the rest of the system

helm/cluster-aws/templates/_helpers.tpl
calvix: ATM that is not supported with the current CAPI image builder as it doesn't provide specific flatcar version in the name or even pin the version, I think it just pick the latest version from the stable channel each time it run

the AMI name looks like this `capa-ami-flatcar-stable-1.22.10-00-gs`

I will create a note  to discuss this, but nothing we can do now

fiunchinho (approved):

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -148,19 +139,79 @@ room for such suffix.
...
+- path: /etc/sysctl.d/hardening.conf
+  permissions: "0644"
   encoding: base64
-  content: {{ $.Files.Get "files/opt/set-aws-ntp.sh" | b64enc }}
+  content: {{ $.Files.Get "files/etc/sysctl.d/hardening.conf" | b64enc }}
 {{- end -}}
 
-{{- define "sshPreKubeadmCommands" -}}
-- systemctl restart sshd
+{{- define "diskStorageConfig" -}}
```
primeroz, line 153: if you have not yet you might want to double check that the filesystems are actually being mounted 

in capz we hit a bug with the way that the `systemd mount units` are generated  see 
* https://github.com/kubernetes-sigs/cluster-api/issues/7679 
* https://github.com/giantswarm/cluster-azure/blob/main/helm/cluster-azure/templates/_kcp.tpl#L72-L88

where essentially we needed to have both `kubeadmConfigSpec.diskSetup` and `ignition.containerLinuxConfig.additionalConfig.storage` to actually get the right volume mounted on the right path

calvix: pretty sure the filesystems are mounted, but I created custom systemd units for the mounts maybe that's the difference?
https://github.com/giantswarm/cluster-aws/pull/322/files#diff-72301810e9b39353e074da2c29eadb7e61d61ca28007476794b2a07103336500R174-R211

AndiDog (approved):

PR #334, opened 2023-07-05
teleport.service: Secure SSH access via Teleport

tuladhar:

Towards https://github.com/giantswarm/roadmap/issues/2193 & https://github.com/giantswarm/roadmap/issues/2786

- Teleport

### Checklist

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites


helm/cluster-aws/files/opt/install-teleport.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
kopiczko, line 1:
```suggestion
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail
```

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -150,6 +150,26 @@ room for such suffix.
...
+  permissions: "0644"
+  encoding: base64
+  content: {{ $.Files.Get "files/opt/install-teleport.sh" | b64enc }}
+{{- end -}}
+
+{{- define "filesConfig" -}}
+- path: "/opt/teleport-v13.1.5-linux-amd64-bin.tar.gz"
+  contents:
+    remote:
+      url: "https://cdn.teleport.dev/teleport-v13.1.5-linux-amd64-bin.tar.gz"
```
kopiczko, line 170: Why not downloading it in `install-teleport.sh`?

alex-dabija, line 170: We probably also need a checksum or signature check before unpacking the tarball and running the install script in order to ensure that the tarball wasn't changed.

tuladhar: Downloading via script doesn't seem to work somehow, I will try it again and share the logs. Thats' why this option was selected.

helm/cluster-aws/files/opt/teleport-installer.sh
```diff
@@ -0,0 +1,125 @@
...
+  is_teleport_installed
+}
+
+if [ $# -ge 1 ] && [ -n "$1" ]; then
+  TELEPORT_VERSION=$1
+else
+  echo "ERROR: Please provide the version you want to install (e.g., 10.1.9)."
+  exit 1
+fi
+is_teleport_installed || install_teleport
```
calvix, line 125: Isn't there a better way how to install or run teleport on the node rather than bash script?

What about a container? this can very easily error or fail  and bash do not have any good way how to retry in such cases and it will be impossible to debug

tuladhar: Unfortunately, no 😞  Spinning up a teleport container was successful but we weren't able to get SSH access via that teleport container, more details here: https://gigantic.slack.com/archives/C053JHJC99Q/p1688643070060619

calvix: I personally do not like this, it's often an important tool in case a node fails and this has no retries or recovery system.

If container is not an option we should rather bake it into the AMI building process rather than download it for every single machine every time it runs.

tuladhar: Totally agree, we should bake it in AMI.

helm/cluster-aws/values.yaml
```diff
@@ -39,6 +39,9 @@ connectivity:
         - availabilityZone: c
           cidr: 10.0.192.0/18
       isPublic: false
+  teleport:
+    proxy: test.teleport.giantswarm.io:443
```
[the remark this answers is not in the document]
tuladhar: It's setup in GiantSwarm AWS account.

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -150,6 +150,27 @@ room for such suffix.
   content: {{ $.Files.Get "files/etc/sysctl.d/hardening.conf" | b64enc }}
 {{- end -}}
 
+{{- define "teleportFiles" -}}
+- path: /etc/teleport-join-token
+  permissions: "0644"
+  contentFrom:
+    secret:
+      name: {{ include "resource.default.name" $ }}-teleport-join-token
```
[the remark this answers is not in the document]
tuladhar: This secret is created by teleport-operator in cluster namespace. We

helm/cluster-aws/files/opt/teleport-installer.sh
fiunchinho: What's the resolution / decision here? Are you going to bake it into the AMI @tuladhar ?

tuladhar: What are your thoughts on this @kopiczko @alex-dabija  Should we tackle it now or later? 

I need help with getting rough idea on how much effort we need to put to create our own custom Flatcar AMI, setup pipeline, etc.

calvix: We already buid our own flatcar AMIs and have working pipeline with capi image builder

so the task would be extend that so it also downloads and prepare teleport binary
https://github.com/giantswarm/capi-image-builder/blob/main/helm/capi-image-builder/templates/pipelines/capa.yaml

alex-dabija: I'm for having the agent in the AMI.

tuladhar: Awesome, collaborating with tinkerers on this! Thanks for the pointer @calvix

helm/cluster-aws/files/etc/teleport.yaml
```diff
@@ -0,0 +1,40 @@
...
+    command: [hostname]
+    period: 24h0m0s
+  - name: arch
+    command: [uname, -m]
+    period: 24h0m0s
+  - name: role
+    command: [/opt/teleport-node-role.sh]
+    period: 1m0s
+  labels:
+    mc: {{ .Values.managementCluster }}
```
kopiczko, line 30: This looks sort of strange that we are using an abbreviation only here. Any reason it isn't `managementCluster:`?

helm/cluster-aws/files/opt/teleport-node-role.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if pgrep "kubelet" > /dev/null; then
+    if pgrep "kube-apiserver" > /dev/null; then
+        echo "control-plane"
+    else
+        echo "worker"
+    fi
+else
+    echo ""
+fi
```
kopiczko, line 15: That looks fragile to me. What if this is being run while kubelet (or apiserver) isn't running? I'd rather base it on the presence of the `kubelet` service and the static apiserver pod manifest

helm/cluster-aws/templates/_helpers.tpl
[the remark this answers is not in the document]
kopiczko: Maybe it would be good to add a comment

helm/cluster-aws/files/opt/teleport-node-role.sh
tuladhar: This script is periodically ran by teleport, so when kubelet is not running it will use "" until next time script runs again.

https://github.com/giantswarm/cluster-aws/blob/bf8c02d3379a70c5e85bea0a3baa73e0df9c7854/helm/cluster-aws/files/etc/teleport.yaml#L26-L28

But, definitely, we can use presence of kubelet service, and static api pod manifest as well, which I think is more quicker as we're only checking files and not process.

tuladhar: Done.

helm/cluster-aws/templates/_helpers.tpl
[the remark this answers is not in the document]
tuladhar: Added the comment.

helm/cluster-aws/files/opt/teleport-installer.sh
tuladhar: Discussion thread with Marcus
- https://gigantic.slack.com/archives/C04KA9DBMB2/p1690790585949089

helm/cluster-aws/files/opt/teleport-installer.sh
```diff
@@ -15,7 +15,7 @@ download() {
 
 # if teleport is already installed, exit
 is_teleport_installed() {
-  if type teleport &>/dev/null; then
+  if type /opt/bin/teleport &>/dev/null; then
```
AndiDog, line 18: Now that the binary is part of the AMI, shouldn't this whole script go away, again reducing the size of the user data (which is limited to 16k)?

CHANGELOG.md
```diff
@@ -114,7 +118,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Set value for `controller-manager` `terminated-pod-gc-threshold` to `125` ( consistent with vintage )
+- Set value for `controller-manager` `terminated-pod-gc-threshold` to `125` ( consistent with vintage )
```
AndiDog, line 121: This is extraneous. You could configure the editor to auto-trim whitespace on save.

helm/cluster-aws/files/etc/teleport.yaml
```diff
@@ -0,0 +1,32 @@
...
+    token_name: /etc/teleport-join-token
+    method: token
+  proxy_server: {{ .Values.internal.teleport.proxyAddr }}
+  log:
+    output: stderr
+auth_service:
+  enabled: "no"
+ssh_service:
+  enabled: "yes"
+  commands:
```
AndiDog, line 14: What are these? We do get full SSH access, so do these help in addition? I guess they provide the "selector" values?! If so, please document the meaning here.

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -145,6 +151,27 @@ room for such suffix.
   content: {{ $.Files.Get "files/etc/sysctl.d/hardening.conf" | b64enc }}
 {{- end -}}
 
+{{/*
+The secret `-teleport-join-token` is created by the teleport-operator in cluster namespace
```
AndiDog, line 155: What's the "cluster namespace"?

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -145,6 +151,27 @@ room for such suffix.
...
+and is used to join the node to the teleport cluster.
+*/}}
+{{- define "teleportFiles" -}}
+- path: /etc/teleport-join-token
+  permissions: "0644"
+  contentFrom:
+    secret:
+      name: {{ include "resource.default.name" $ }}-teleport-join-token
```
AndiDog, line 163: If this comes from an operator, what about MC creation? Are we adding this token pre-creation to mc-bootstrap? It seems a chicken-and-egg prerequisite before CAPA would reconcile the MC's `AWSCluster`, if I'm not mistaken.

helm/cluster-aws/values.schema.json
```diff
@@ -653,6 +653,27 @@
...
                     }
+                },
+                "teleport": {
+                    "type": "object",
+                    "title": "Teleport",
+                    "properties": {
+                        "enabled": {
+                            "type": "boolean",
+                            "title": "Enable teleport",
+                            "default": false
```
AndiDog, line 664: Why not turn it on?

helm/cluster-aws/templates/_helpers.tpl
calvix: if the teleport operator is deployed and working  into the kind cluster in mc bootstrap phase it should be fine

We do a similar thing with `irsa-operator` and `encryption-provider-operator`

helm/cluster-aws/values.schema.json
calvix: +1 i would say we should enable by default

calvix:
I think you missed to run `make generate-docs` for the value update

helm/cluster-aws/files/etc/teleport.yaml
tuladhar: I am not sure what you meant by "selector" values. But, here we have configure teleport to run in SSH service mode only to provide SSH access to the nodes via teleport. For example, the Kubernetes access is handled by [teleport-kube-agent](https://github.com/giantswarm/teleport-kube-agent-app) app.

helm/cluster-aws/templates/_helpers.tpl
tuladhar: It's the namespace where MC or WC cluster is deployed in.

helm/cluster-aws/templates/_helpers.tpl
tuladhar: `teleport-operator` will be deploy from mc-bootstrap during MC creation, and the teleport join token secret will exists before we pivot MC from kind cluster to AWS, so we avoid chicken-egg problem. Does that make sense?

helm/cluster-aws/values.schema.json
tuladhar: Currently we only have test teleport cluster and working on production teleport cluster, so not sure if it makes sense to turn it on.

helm/cluster-aws/templates/_helpers.tpl
tuladhar: mc-bootstrap PR - https://github.com/giantswarm/mc-bootstrap/pull/676

calvix (approved):
lgtm

kopiczko:
In general it looks good. But we will have to put this in all provider charts. To make the task easier and maybe even use `vendir` to sync it from a separate repo I'd try to use a subchart for the `define` statements. See https://github.com/helm/helm/pull/9957

tuladhar:
> In general it looks good. But we will have to put this in all provider charts. To make the task easier and maybe even use `vendir` to sync it from a separate repo I'd try to use a subchart for the `define` statements. See [helm/helm#9957](https://github.com/helm/helm/pull/9957)

Team Turtles is working on a provider-independent cluster chart where almost all provider-independent resources will be. So whatever provider-independent stuff (subchart reference or resources themselves) we put into cluster-aws and other cluster apps, will move it to cluster chart.

AndiDog (approved):

kopiczko (approved):

PR #341, opened 2023-07-12
Support creating `CiliumNetworkPolicy` manifests that allow egress requests to DNS and conditionally the proxy host, covering only the `giantswarm` and `kube-system` namespaces

AndiDog:

Towards https://github.com/giantswarm/giantswarm/issues/27266

For CAPA proxy-private MCs (such as `goat`), we need to allow critical applications such as our operators to reach the proxy. The proxy is defined by hostname, e.g. `proxy.goatproxy.gaws.gigantic.internal` or `internal-foobar.eu-north-1.elb.amazonaws.com`. Cilium offers L7-level allow rules: `toFQDNs`. To make that work, Cilium needs to capture DNS responses to cache the mapping from hostname to IPs, of which it will manage L3 allow-these-IPs rules. Since we don't want to copy-paste a proxy-related network policy solution into every operator app, it makes sense to put it into the cluster app (here: `cluster-aws`), since we deploy Cilium and proxy settings from here.

As Cilium's [default enforcement mode](https://docs.cilium.io/en/latest/security/policy/intro/) turns all pods into default-deny-(egress/ingress) once a pod is selected by any policy, even a cluster-wide policy, we reduce the scope of the new network policies down to critical namespaces. Together with https://github.com/giantswarm/giantswarm/issues/23014, all operators should work fine.

I tested on `golem` (only the DNS policy will be deployed) and on `goat` (also the allow-egress-to-proxy policy gets deployed). Enabling `connectivity.createCiliumNetworkPolicies=true` (feature of _this_ PR, to be enabled on all MCs by default) will be coordinated together with CAPA network policies (https://github.com/giantswarm/cluster-api-provider-aws-app/pull/173) and ultimately the introduction of default-deny-all policies for egress _and_ ingress.

Request for reviewers: is the documentation enough? Should we explain more about its use, which is meant for MCs?

### Checklist

### Trigger e2e tests

/test create
/test upgrade
/run cluster-test-suites

### DIFF
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -205,6 +205,12 @@
                         ]
                     }
                 },
+                "createCiliumNetworkPolicies": {
+                    "type": "boolean",
+                    "title": "Create CiliumNetworkPolicy manifests for DNS and proxy access in critical namespaces",
+                    "description": "For the namespaces `giantswarm` and `kube-system`, this creates a network policy allowing DNS access. If `proxy.enabled` is true, egress to the proxy is allowed as well, but for now this only supports an FQDN, not an IP address, inside the `httpProxy`/`httpsProxy` URLs.",
+                    "default": false
+                },
                 "dns": {
                     "type": "object",
                     "title": "DNS",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support creating `CiliumNetworkPolicy` manifests that allow egress requests to DNS and conditionally the proxy host, covering only the `giantswarm` and `kube-system` namespaces
+
 ## [0.36.1] - 2023-07-11
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


AndiDog:
Marking as draft for now since this doesn't really make sense for WCs as the policies need to be deployed within the WC.

fiunchinho:
If this only applies to MCs, what about adding it to `management-clusters-fleet` when running mc-bootstrap?

helm/cluster-aws/values.schema.json
```diff
@@ -205,6 +205,12 @@
                         ]
                     }
                 },
+                "createCiliumNetworkPolicies": {
+                    "type": "boolean",
+                    "title": "Create CiliumNetworkPolicy manifests for DNS and proxy access in critical namespaces",
```
marians, line 210: This title is a bit long. If needed, information can be moved into the description.

```suggestion
                    "title": "Create CiliumNetworkPolicy manifests",
```

AndiDog:
> If this only applies to MCs, what about adding it to `management-clusters-fleet` when running mc-bootstrap?

That's one good idea, indeed.

But a customer might be interested in the solution particularly for WCs that use a proxy. In such a case we need to deploy the policies there. For that, we basically need a separate app repo (e.g. `proxy-private-policies-app`) and could then 1) reference it via `HelmRelease` from here (`cluster-aws` installs a child chart) or 2) make it completely separate and thereby not clutter the `cluster-aws` repo (which means we have to add the app installation to mc-bootstrap).

Let's discuss in the standup.

AndiDog:
We discussed that it makes sense to put the policies into `cilium-app` since chart-operator in WCs definitely needs egress allowance.

PR #343, opened 2023-07-13
Add value to configure control plane load balancer allow list

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/2351

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add teleport.service: Secure SSH access via Teleport.
-- Add `controlPlane.loadBalancerIngressRules` to configure control plane load balancer ingress rules.
+- Add `controlPlane.loadBalancerIngressAllowCidrBlocks` to configure control plane load balancer ingress rules.
 
 ## [0.45.0] - 2023-10-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add teleport.service: Secure SSH access via Teleport
+- Add teleport.service: Secure SSH access via Teleport.
+- Add `controlPlane.loadBalancerIngressRules` to configure control plane load balancer ingress rules.
 
 ## [0.45.0] - 2023-10-04
 
--- a/helm/cluster-aws/README.md
+++ b/helm/cluster-aws/README.md
@@ -100,8 +100,8 @@ Properties within the `.controlPlane` top-level object
 | `controlPlane.etcdVolumeSizeGB` | **Etcd volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
 | `controlPlane.instanceType` | **EC2 instance type**|**Type:** `string`<br/>**Default:** `"r6i.xlarge"`|
 | `controlPlane.kubeletVolumeSizeGB` | **Kubelet volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
-| `controlPlane.loadBalancerIngressRules` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
-| `controlPlane.loadBalancerIngressRules[*]` | **Address range**|**Type:** `string`<br/>|
+| `controlPlane.loadBalancerIngressAllowCidrBlocks` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
+| `controlPlane.loadBalancerIngressAllowCidrBlocks[*]` | **Address range**|**Type:** `string`<br/>|
 | `controlPlane.machineHealthCheck` | **Machine health check**|**Type:** `object`<br/>|
 | `controlPlane.machineHealthCheck.enabled` | **Enable**|**Type:** `boolean`<br/>**Default:** `true`|
 | `controlPlane.machineHealthCheck.maxUnhealthy` | **Maximum unhealthy nodes**|**Type:** `string`<br/>**Example:** `"40%"`<br/>**Default:** `"40%"`|
--- a/helm/cluster-aws/README.md
+++ b/helm/cluster-aws/README.md
@@ -100,7 +100,7 @@ Properties within the `.controlPlane` top-level object
 | `controlPlane.etcdVolumeSizeGB` | **Etcd volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
 | `controlPlane.instanceType` | **EC2 instance type**|**Type:** `string`<br/>**Default:** `"r6i.xlarge"`|
 | `controlPlane.kubeletVolumeSizeGB` | **Kubelet volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
-| `controlPlane.loadBalancerIngressAllowCidrBlocks` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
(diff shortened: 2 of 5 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -478,6 +478,15 @@
             "type": "object",
             "title": "Control plane",
             "properties": {
+                "allowList": {
+                    "type": "array",
+                    "title": "Load Balancer Allow list",
```
marians, line 483:
```suggestion
                    "title": "Load balancer allow list",
```

helm/cluster-aws/values.schema.json
```diff
@@ -472,6 +472,18 @@
             "type": "object",
             "title": "Control plane",
             "properties": {
+                "allowList": {
+                    "type": "array",
+                    "title": "Load balancer allow list",
+                    "description": "IPs that are allowed to connect to the control plane load balancer.",
```
marians, line 478: Aligning this with other CIDR properties and their descriptions:

```suggestion
                    "description": "IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation.",
```

helm/cluster-aws/values.schema.json
```diff
@@ -472,6 +472,18 @@
...
+                    "items": {
+                        "type": "string",
+                        "title": "CIDR"
```
marians, line 481:
```suggestion
                        "title": "Address range"
```

marians:
I think schemadocs needs updating, too

fiunchinho:
The e2e tests on this PR will fail until CAPA `v2.2.3` release is deployed on `grizzly`.

helm/cluster-aws/values.schema.json
```diff
@@ -472,6 +472,18 @@
             "type": "object",
             "title": "Control plane",
             "properties": {
+                "allowList": {
+                    "type": "array",
+                    "title": "Load balancer allow list",
+                    "description": "IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by GiantSwarm so that the cluster can still be managed.",
```
marians, line 478: Brand police here 🚨 

```suggestion
                    "description": "IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.",
```

helm/cluster-aws/README.md
```diff
@@ -83,6 +83,8 @@ Properties within the `.controlPlane` top-level object
 
 | **Property** | **Description** | **More Details** |
 | :----------- | :-------------- | :--------------- |
+| `controlPlane.allowList` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by GiantSwarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
```
marians, line 86:
```suggestion
| `controlPlane.allowList` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
```

helm/cluster-aws/templates/_aws_cluster.tpl
```diff
@@ -32,6 +32,15 @@ spec:
...
   controlPlaneLoadBalancer:
     scheme: {{ if (eq .Values.controlPlane.apiMode "public") }}internet-facing{{ else }}internal{{ end }}
+    {{- if .Values.controlPlane.allowList }}
+    ingressRules:
+    - description: "Kubernetes API"
+      protocol: tcp
+      fromPort: 6443
+      toPort: 6443
+      cidrBlocks:
+      {{- toYaml .Values.controlPlane.allowList | nindent 6 }}
```
AndiDog, line 42: Indentation needs to match the existing list item for the Kubernetes API, so shouldn't it be `nindent 4`?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `controlPlane.allowList` to configure control plane load balancer ingress rules.
```
AndiDog, line 12: The name is quite unclear. What about `controlPlane.loadBalancerIngressRules` which matches CAPA naming? Also, it looks like the default value `["0.0.0.0/0"]` shouldn't work because `[]IngressRule` is expected, not `[]string`.

The description shouldn't say anything about IPv4. The `IngressRule` type supports IPv6 as well.

Also, should we move this into `.Values.providerSpecific`?

AndiDog (changes requested):

helm/cluster-aws/templates/_aws_cluster.tpl
fiunchinho: If I use `nindent 4` it would look like this, and it wouldn't work

```
- description: "Kubernetes API"
      protocol: tcp
      fromPort: 6443
      toPort: 6443
      cidrBlocks:
    - 1.2.3.4/32
```

CHANGELOG.md
fiunchinho: Regarding `IPv4`, we are currently only setting `cidrBlocks` field, which is IPv4. I don't know if we support clusters using IPv6 at this point.

AndiDog: I was confused by

```
cidrBlocks:
  {{- toYaml .Values.controlPlane.loadBalancerIngressRules | ... }}
```

because the name `controlPlane.loadBalancerIngressRules` reads as if users can pass additional rules. Therefore I commented as if that was the goal.

However, what you currently implement is rather something like `controlPlane.loadBalancerIngressAllowCidrBlocks`, right? I'm fine to only allow configuring IPv4 blocks, defaulting to `0.0.0.0/0`, but then we need to give the value a good name. We can still migrate to supporting the configuration of full rules _later_.

helm/cluster-aws/README.md
```diff
@@ -100,6 +100,8 @@ Properties within the `.controlPlane` top-level object
 | `controlPlane.etcdVolumeSizeGB` | **Etcd volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
 | `controlPlane.instanceType` | **EC2 instance type**|**Type:** `string`<br/>**Default:** `"r6i.xlarge"`|
 | `controlPlane.kubeletVolumeSizeGB` | **Kubelet volume size (GB)**|**Type:** `integer`<br/>**Default:** `100`|
+| `controlPlane.loadBalancerIngressAllowCidrBlocks` | **Load balancer allow list** - IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.|**Type:** `array`<br/>**Default:** `["0.0.0.0/0"]`|
```
AndiDog, line 103: At best, customers shouldn't need to take care of adding MC gateway IPs. Can we `{{ lookup [...] "AWSCluster" .Values.managementCluster }}` (pseudocode) the MC and check its `AWSCluster.status.networkStatus.natGatewaysIPs`?

helm/cluster-aws/values.schema.json
```diff
@@ -585,6 +585,18 @@
...
                 },
+                "loadBalancerIngressAllowCidrBlocks": {
+                    "type": "array",
+                    "title": "Load balancer allow list",
+                    "description": "IPv4 address ranges that are allowed to connect to the control plane load balancer, in CIDR notation. When setting this field, remember to add the Management cluster Nat Gateway IPs provided by Giant Swarm so that the cluster can still be managed.",
+                    "items": {
+                        "type": "string",
+                        "title": "Address range"
+                    },
+                    "default": [
```
AndiDog, line 596: We could also default to leaving this off, as before

helm/cluster-aws/README.md
fiunchinho: We discussed on Slack to avoid using `lookup` and just document how to get the IPs.

helm/cluster-aws/values.schema.json
fiunchinho: Removed

whites11 (approved):

helm/cluster-aws/values.schema.json
```diff
@@ -585,6 +585,15 @@
                     "title": "Kubelet volume size (GB)",
                     "default": 100
                 },
+                "loadBalancerIngressAllowCidrBlocks": {
```
calvix, line 588: I personally do not like that adding MC nat gateways is a manual step

calvix: this will be often forgotten by customer and we will get lot of pages because of it

fiunchinho: We could automate it in `kubectl-gs template cluster`

helm/cluster-aws/templates/_aws_cluster.tpl
```diff
@@ -32,6 +32,17 @@ spec:
...
     scheme: {{ if (eq .Values.controlPlane.apiMode "public") }}internet-facing{{ else }}internal{{ end }}
+    {{- if .Values.controlPlane.loadBalancerIngressAllowCidrBlocks }}
+    ingressRules:
+    - description: "Kubernetes API"
+      protocol: tcp
+      fromPort: 6443
+      toPort: 6443
+      cidrBlocks:
+      - 185.102.95.187/32
+      - 95.179.153.65/32
```
AndiDog, line 43: There must be a comment for these, and at best a link to where those come from

calvix (approved):

AndiDog (approved):

PR #354, opened 2023-08-22
vintage-migration-changes

calvix:
additional field necessary for migration from the vintage cluster
ref issue: https://github.com/giantswarm/giantswarm/issues/27910

the values are filled like this
https://github.com/giantswarm/capi-migration-cli/blob/main/pkg/migrator/capi_app.go#L79-L148

### DIFF
--- a/helm/cluster-aws/templates/_control_plane.tpl
+++ b/helm/cluster-aws/templates/_control_plane.tpl
@@ -30,9 +30,9 @@ template:
       size: {{ .Values.controlPlane.rootVolumeSizeGB }}
       type: gp3
     iamInstanceProfile: control-plane-{{ include "resource.default.name" $ }}
-    {{- if .Values.controlPlane.additionalSecurityGroupID }}
+    {{- if .Values.internal.migration.controlPlaneAdditionalSecurityGroupID }}
     additionalSecurityGroups:
-    - id: {{ .Values.controlPlane.additionalSecurityGroupID }}
+    - id: {{ .Values.internal.migration.controlPlaneAdditionalSecurityGroupID }}
     {{- end }}
     sshKeyName: ""
     subnet:
--- a/helm/cluster-aws/templates/_control_plane.tpl
+++ b/helm/cluster-aws/templates/_control_plane.tpl
@@ -93,6 +93,9 @@ spec:
         certSANs:
           - "api.{{ include "resource.default.name" $ }}.{{ required "The baseDomain value is required" .Values.baseDomain }}"
           - 127.0.0.1
+          {{- if .Values.controlPlane.apiExtraCertSANs -}}
+          {{- toYaml .Values.controlPlane.apiExtraCertSANs | nindent 10 }}
+          {{- end }}
         extraArgs:
           cloud-provider: external
           service-account-issuer: "irsa.{{ include "resource.default.name" $ }}.{{ required "The baseDomain value is required" .Values.baseDomain }}"
@@ -122,6 +125,9 @@ spec:
           service-account-lookup: "true"
           tls-cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
           service-cluster-ip-range: {{ .Values.connectivity.network.services.cidrBlocks | first }}
+          {{- if .Values.controlPlane.apiExtraArgs -}}
+          {{- toYaml .Values.controlPlane.apiExtraArgs | nindent 10 }}
+          {{- end }}
         extraVolumes:
         - name: auditlog
           hostPath: /var/log/apiserver
@@ -157,6 +163,9 @@ spec:
           extraArgs:
             listen-metrics-urls: "http://0.0.0.0:2381"
             quota-backend-bytes: "8589934592"
+            {{- if .Values.controlPlane.apiExtraArgs -}}
+            {{- toYaml .Values.internal.migration.etcdExtraArgs | nindent 12 }}
+            {{- end }}
       networking:
         serviceSubnet: {{ join "," .Values.connectivity.network.services.cidrBlocks }}
     files:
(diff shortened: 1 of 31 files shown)


helm/cluster-aws/values.schema.json
```diff
@@ -621,6 +682,70 @@
...
+                "migration": {
+                    "type": "object",
+                    "title": "Migration values",
+                    "description": "Section used for migration of cluster from vintage to CAPI",
+                    "properties": {
+                        "apiBindPort": {
+                            "type": "integer",
+                            "title": "Kubernetes API bind port",
+                            "description": "Kubernetes API bind port used for kube api pod",
+                            "default": 6443
```
calvix, line 694: vintage listen on 443 and in the migration step we need to also set this to port 443, could be removed later

helm/cluster-aws/values.schema.json
```diff
@@ -621,6 +682,70 @@
...
+                        },
+                        "controlPlaneExtraFiles": {
```
calvix, line 696: mostly for adding migration script for etcd

calvix:
the example how the new values are filled can be found here [capi_app](https://github.com/giantswarm/capi-migration-cli/blob/main/pkg/migrator/capi_app.go)

helm/cluster-aws/templates/_control_plane.tpl
```diff
@@ -30,6 +30,10 @@ template:
       size: {{ .Values.controlPlane.rootVolumeSizeGB }}
       type: gp3
     iamInstanceProfile: control-plane-{{ include "resource.default.name" $ }}
+    {{- if .Values.controlPlane.additionalSecurityGroupID }}
```
AndiDog, line 33: Let's make this plural and probably use the same type as CAPA i.e.

```
# values.yaml
controlPlane:
  additionalSecurityGroups:
    - id: sg-foobar
```

helm/cluster-aws/templates/_control_plane.tpl
```diff
@@ -167,13 +180,23 @@ spec:
     {{- if .Values.connectivity.proxy.enabled }}{{- include "proxyFiles" . | nindent 4 }}{{- end }}
     {{- include "kubernetesFiles" . | nindent 4 }}
     {{- include "registryFiles" . | nindent 4 }}
+    {{- if .Values.internal.migration.controlPlaneExtraFiles }}
+    {{- range $file := .Values.internal.migration.controlPlaneExtraFiles }}
```
AndiDog, line 184: Let's use the same type in `internal.migration.controlPlaneExtraFiles` so we don't have to rewrite from one structure to the other and remain flexible:

```
# values.yaml example
internal:
  migration:
    controlPlaneExtraFiles:
      - path: {{ $file.path }}
        permissions: "0644"
        contentFrom:
          secret:
           name: mysecret
           key: mypath
```

helm/cluster-aws/templates/_control_plane.tpl
```diff
@@ -210,12 +236,18 @@ spec:
...
-    {{- include "flatcarKubeadmPreCommands" . | nindent 4 }}
     {{- include "sshPreKubeadmCommands" . | nindent 4 }}
+    {{- if .Values.internal.migration.controlPlanePreKubeadmCommands -}}
+    {{- toYaml .Values.internal.migration.controlPlanePreKubeadmCommands | nindent 4 }}
+    {{- end }}
+    {{- include "flatcarKubeadmPreCommands" . | nindent 4 }}
     {{- if .Values.connectivity.proxy.enabled }}{{- include "proxyCommand" $ | nindent 4 }}{{- end }}
     postKubeadmCommands:
-    {{- include "kubeletConfigPostKubeadmCommands" . | nindent 4 }}
+    {{- include "kubeletConfigPostKubeadmCommands" . | nindent 4 }}
```
AndiDog, line 246: Please trim whitespace

helm/cluster-aws/templates/_helpers.tpl
```diff
@@ -221,6 +221,10 @@ room for such suffix.
 - name: giantswarm
   groups: sudo
   sudo: ALL=(ALL) NOPASSWD:ALL
+- name: calvix
```
AndiDog, line 224: To be removed, I guess

helm/cluster-aws/templates/_machine_pools.tpl
```diff
@@ -66,6 +66,10 @@ spec:
       size: {{ $value.rootVolumeSizeGB | default 300 }}
       type: gp3
     sshKeyName: ""
+    {{- if $value.additionalSecurityGroupID }}
```
AndiDog, line 69: Same as above, let's use plural and CAPA's structure, and not convert from ID string to their structure

helm/cluster-aws/values.schema.json
```diff
@@ -295,6 +305,11 @@
                             "title": "VPC subnet",
                             "description": "IPv4 address range to assign to this cluster's VPC, in CIDR notation.",
                             "default": "10.0.0.0/16"
+                        },
+                        "vpcId": {
+                            "type": "string",
+                            "title": "VPC id",
+                            "description": "ID of the VPC, where the cluster will be deployed."
```
fiunchinho, line 312: I'm guessing when we specify the vpc ID, the VPC is not created by CAPA? If that's the case, I'd make it clear in the description.

helm/cluster-aws/values.schema.json
```diff
@@ -367,10 +382,25 @@
                                     }
                                 }
                             },
+                            "id": {
+                                "type": "string",
+                                "title": "ID of the subnet",
+                                "description": "ID of the subnets, used inc ase we want to reuse already existing subnet."
```
fiunchinho, line 388:
```suggestion
                                "description": "ID of an existing subnet. When set, this subnet will be used instead of creating a new one."
```

helm/cluster-aws/values.schema.json
```diff
@@ -614,6 +675,70 @@
...
+                        "controlPlanePreKubeadmCommands": {
+                            "type": "array",
+                            "title": "Control Plane Pre Kubeadm Commands",
+                            "description": "Additional Pre-Kubeadm Commands executed on the control plane node.",
+                            "items": {
+                                "type": "string",
+                                "title": "command"
+                            }
+                        },
+                        "etcdExtraArgs": {
```
fiunchinho, line 730: Do we need to support different etcd configurations? I was thinking if we need this value at all. Couldn't we just use whatever flags we need in `clusterConfiguration.etcd.local.extraArgs` and avoid making it configurable?

helm/cluster-aws/templates/_helpers.tpl
calvix: Yeah, just for testing

helm/cluster-aws/values.schema.json
calvix: no, because this extra arguments are only needed for the migration, so it cannot be set globally for all clusters

anything under `.internal` should not be used by customers and anything under `migration` should be possible to remove after migration

calvix:
once i get LGTM i will remove the custom user and merge but until then i need the user for testing purposes

fiunchinho (approved):

PR #355, opened 2023-08-23
Tolerate CAPI taints on uninitialized nodes

fiunchinho:

`cilium` helm release takes very long to become ready because cilium ui and cilium relay pods can't be scheduled on CAPI uninitialized nodes. I think it's safe to add these tolerations, because CAPI initialization of nodes is just syncing labels.

### Checklist


calvix (approved):
would it be useful for all helmRelease based apps?

fiunchinho:
> would it be useful for all helmRelease based apps?

I didn't find it to be an issue for any of the others, so to keep things simple, I'd keep it only for cilium for now.

PR #364, opened 2023-08-29
Add schema to service account issuer URI

mcharriere:

The `service-account-issuer` arg requires a URI, which should include the `schema`.

> Identifier  of the service account token issuer. The issuer will assert this  identifier in "iss" claim of issued tokens. This value is a string or  URI. If this option is not a valid URI per the OpenID Discovery 1.0  spec, the ServiceAccountIssuerDiscovery feature will remain disabled,  even if the feature gate is set to true. It is highly recommended that  this value comply with the OpenID spec:  https://openid.net/specs/openid-connect-discovery-1_0.html. In practice,  this means that service-account-issuer must be an https URL. It is also  highly recommended that this URL be capable of serving OpenID discovery  documents at {service-account-issuer}/.well-known/openid-configuration.  When this flag is specified multiple times, the first is used to  generate tokens and all are used to determine which issuers are  accepted.

More details https://gigantic.slack.com/archives/C02GLN41UP2/p1693239516555659

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Add `scheme` to service-account-issuer URI.
+
 ## [0.38.3] - 2023-08-29
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


calvix (approved):

fiunchinho (approved):

AndiDog (approved):

mcharriere:
here we go again... tests are taking for ever 😢

PR #373, opened 2023-09-12
Release v0.39.0

taylorbot:


AndiDog:
Tested via the only change https://github.com/giantswarm/cluster-aws/pull/371, not running the test suite again

PR #377, opened 2023-09-18
Add old issuer URL as second `--service-account-issuer` parameter for API server to avoid a breaking change in v0.38.4

AndiDog:

Fixes https://github.com/giantswarm/giantswarm/issues/28165

In short: use two arguments `--service-account-issuer=https://irsa.[...] --service-account-issuer=irsa.[...]` so old and new service account tokens are allowed during upgrade and until the old tokens expire and get rotated.

I tested this by upgrading from v0.38.3 to `v0.40.0 + my changes`, and the Kubernetes API was only unavailable shortly during rollout of nodes. For comparison, I tried the broken upgrade v0.38.3..v0.38.4 and it gave me 100% `error: You must be logged in to the server (Unauthorized)` errors after the rollout because all existing tokens still had the old (and not accepted) issuer URI without the `https://` prefix.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+⚠️ When upgrading, please use v0.41.0 (_this release_) or newer. See our note on the breaking change in v0.38.4.
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


fiunchinho (approved):
Nice job

PR #380, opened 2023-09-20
remove-support-for-dns-mode-private

calvix:
towartds https://github.com/giantswarm/roadmap/issues/2690

### DIFF
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -241,10 +241,9 @@
                         "mode": {
                             "type": "string",
                             "title": "Mode",
-                            "description": "Whether the Route53 hosted zone of this cluster should be public or private.",
+                            "description": "Whether the Route53 hosted zone of this cluster should be public or private. Only 'public' mode is allowed at the moment",
                             "enum": [
-                                "public",
-                                "private"
+                                "public"
                             ],
                             "default": "public"
                         },
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -225,19 +225,6 @@
                     "type": "object",
                     "title": "DNS",
                     "properties": {
-                        "additionalVpc": {
-                            "type": "array",
-                            "title": "Additional VPCs",
-                            "description": "If DNS mode is 'private', the VPCs specified here will be assigned to the private hosted zone.",
-                            "items": {
-                                "type": "string",
-                                "title": "VPC identifier",
-                                "examples": [
-                                    "vpc-x2aeasd1d"
-                                ],
-                                "pattern": "^vpc-[0-0a-zA-Z]+$"
-                            }
-                        },
                         "mode": {
                             "type": "string",
                             "title": "Mode",
--- a/helm/cluster-aws/values.schema.json
+++ b/helm/cluster-aws/values.schema.json
@@ -241,7 +241,7 @@
                         "mode": {
                             "type": "string",
                             "title": "Mode",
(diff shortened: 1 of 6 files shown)


fiunchinho (changes requested):
I think we can remove this now https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/templates/_aws_cluster.tpl#L12-L16

And also the `additionalVpc` field from the schema and values files.

calvix:
> I think we can remove this now [`master`/helm/cluster-aws/templates/_aws_cluster.tpl#L12-L16](https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/templates/_aws_cluster.tpl?rgh-link-date=2023-09-20T15%3A05%3A13Z#L12-L16)
> 
> And also the `additionalVpc` field from the schema and values files.

removed

helm/cluster-aws/values.schema.json
```diff
@@ -225,26 +225,12 @@
                     "type": "object",
                     "title": "DNS",
                     "properties": {
-                        "additionalVpc": {
```
AndiDog, line 228: You need to run

```
make normalize-schema generate-values generate-docs validate-schema
```

AndiDog (changes requested):

helm/cluster-aws/values.schema.json
calvix: I always run that, otherwise the  `JSON schema validation` CI would fail

fiunchinho (commented):
Actually, would it make sense to remove the whole `dns` block? I don't think we need it anymore, right? https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.schema.json#L224-L265

calvix:
> Actually, would it make sense to remove the whole `dns` block? I don't think we need it anymore, right? [`master`/helm/cluster-aws/values.schema.json#L224-L265](https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.schema.json?rgh-link-date=2023-09-21T08%3A06%3A23Z#L224-L265)

Don't we still use the resolver rules in the product?

helm/cluster-aws/values.schema.json
calvix: Oh i was missing `generate-docs` I wonder why that command is missing from the list in the Readme, just added it

calvix: sorry my bad

fiunchinho:
> > Actually, would it make sense to remove the whole `dns` block? I don't think we need it anymore, right? [`master`/helm/cluster-aws/values.schema.json#L224-L265](https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.schema.json?rgh-link-date=2023-09-21T08%3A06%3A23Z#L224-L265)
> 
> Don't we still use the resolver rules in the product?

I think we only needed it when using private DNS, but if we are dropping support for private dns, I don't think we need it anymore. I may be wrong though.

calvix:
@T-Kukawka DO we wanna keep a feature called Resolver rules in our product or do we wanna get rid of it too? 
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver.html

AndiDog (approved):

T-Kukawka:
Personally i would leave this feature if it does not bring much of overhead on our side, especially for the future to target more 'hybrid' customers where this can come in handy.

fiunchinho (approved):

PR #383, opened 2023-09-26
Remove installation of Cilium policies that allow certain cluster traffic unconditionally

AndiDog:

Towards https://github.com/giantswarm/giantswarm/issues/27266

This is hopefully the last generic allow rule. All operators have their own policies and don't seem to need this generic one anymore. See https://github.com/giantswarm/cilium-app/pull/93 and the [default policy manifests](https://github.com/giantswarm/cilium-app/tree/main/helm/cilium/files/policies) which we're not using anymore with this change.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):

AndiDog:
Test fails due to cert-manager issues. I can't depend on that getting fixed. Since my change was tested manually, I'll merge.

PR #401, opened 2023-10-23
remove faulty regex and remove labels from awsmachinetemplate specs

bdehri:

We introduced regexReplace with this [commit](https://github.com/giantswarm/cluster-aws/commit/c5da63f1813c1459af7ad4e91dec8add308b61bc) to prevent unnecessary node rolling which caused no node rolling at all. Instead of removing respective label via regex, all label related sections were moved to AWSMachineTemplate manifests directly.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


helm/cluster-aws/templates/_bastion.tpl
```diff
@@ -102,9 +96,15 @@ metadata:
...
+  name: {{ include "resource.default.name" $ }}-bastion-{{ include "hash" (dict "data" (include "bastion-awsmachinetemplate-spec" $) "global" .) }}
   namespace: {{ .Release.Namespace }}
-spec: {{ include "bastion-awsmachinetemplate-spec" $ | nindent 2 }}
+spec:
+  template:
+    metadata:
+      labels:
+        cluster.x-k8s.io/role: bastion
+        {{- include "labels.common" $ | nindent 8 }}
+    spec:{{- include "controlplane-awsmachinetemplate-spec" $ | nindent 6 }}
```
AndiDog, line 107: [repeated]

The `{{- ... }}` bit can be moved to the next line, since the `-` strips the whitespace before. That's easier to read and search, and consistent with how we have it elsewhere.

AndiDog (commented):
Good idea, and `xxx-spec` now actually means what it says 😄 Ping me when this is tested and ready for review.

AndiDog (approved):
LGTM if tested with a workload cluster

fiunchinho (approved):

PR #405, opened 2023-10-31
expose-aws-additional-tags

calvix:
towards https://github.com/giantswarm/roadmap/issues/2929

/run cluster-test-suites

### DIFF
--- a/helm/cluster-aws/README.md
+++ b/helm/cluster-aws/README.md
@@ -17,6 +17,8 @@ Properties within the `.providerSpecific` top-level object
 
 | **Property** | **Description** | **More Details** |
 | :----------- | :-------------- | :--------------- |
+| `providerSpecific.additionalResourceTags` | **Additional resource tags** - Additional tags to add to AWS resources created by the cluster.|**Type:** `object`<br/>|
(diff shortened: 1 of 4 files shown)


whites11 (approved):

helm/cluster-aws/README.md
```diff
@@ -17,6 +17,8 @@ Properties within the `.providerSpecific` top-level object
 
 | **Property** | **Description** | **More Details** |
 | :----------- | :-------------- | :--------------- |
+| `providerSpecific.additionalResourceTags` | **Additional resource tags** - Additional tags to add to AWS resources created by the cluster.|**Type:** `object`<br/>|
```
fiunchinho, line 20: Do we know which AWS resources are tagged? I think it would be nice to list them here.

AndiDog (approved):

helm/cluster-aws/README.md
calvix:
```
// AdditionalTags is an optional set of tags to add to AWS resources managed by the AWS provider, in addition to the
	// ones added by default.
```
we do not know as its not written specifically, but it should be all resources that are created by capa controller and support tags

calvix: the goal is that all resources created for the cluster should be tagged

fiunchinho (approved):

PR #408, opened 2023-11-03
cluster-autoscaler

njuettner:

Issue: https://github.com/giantswarm/roadmap/issues/2927

Prepare CR's to allow managing MachinePools by cluster-autoscaler.

Cluster-autoscaler watches ASG's on two tags 
*  sigs.k8s.io/cluster-api-provider-aws/cluster/CLUSTER_ID <- already an ASG tag
*  k8s.io/cluster-autoscaler/enabled: "true" <- needs to be created by capa-controller because it's not present yet.

I tried to make an overview how it works:
image

* capa-controller will only adjust the min and max size for MachinePools in case cluster owner modifies AWSMachinePool CR and updating the status of the current replicas
* cluster-autoscaler will take ownership of scaling in and out according to it's calculation in the workload cluster.

Once released we can manage scaling for MachinePools by cluster-autoscaler.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow cluster-autoscaler handling MachinePools.
+- Add additional tag for cluster autoscaler to MachinePool ASGs.
+
 ## [0.46.1] - 2023-11-01
 
 ### Added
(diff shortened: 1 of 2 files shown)


calvix (approved):

njuettner:
Same here: I don't know why but the cluster (CAPA Upgrade Suite) cannot be deleted in time but I doubt it has something to do with cluster-autoscaler

fiunchinho (approved):

PR #411, opened 2023-11-09
Add `global.metadata.preventDeletion` to add the deletion prevention label to cluster resources

AndiDog:
Towards https://github.com/giantswarm/giantswarm/issues/28310. Requires https://github.com/giantswarm/schemalint/pull/115.

Add the label to objects that shouldn't be deletable. Since Helm would try to delete the objects not marked with `"helm.sh/resource-policy": keep`, `Cluster` is the most important parent object to protect. Other than those, I added the label to reasonable objects such as `AWSCluster` (which correlates to the AWS infrastructure), but not to objects whose name contains a hash as they need to be deletable by Helm on upgrade (= when the hash in the name probably changes).

I manually tested cluster creation, and also upgrade with manifest changes.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `global.metadata.preventDeletion` to add the [deletion prevention label](https://docs.giantswarm.io/advanced/deletion-prevention/) to cluster resources
+
 ## [0.47.0] - 2023-11-07
 
 ### Added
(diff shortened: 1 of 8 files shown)


fiunchinho:
Should `MachinePools` and `AWSMachinePools` also contain the label?

calvix (approved):

AndiDog:
> Should `MachinePools` and `AWSMachinePools` also contain the label?

I thought to rather not because pools are customer-defined, so if `.Values.nodePools` changes, it may lead to non-obvious problems because the old objects cannot be deleted.

PR #413, opened 2023-11-11
Move Values.metadata to Values.global.metadata

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/2954

We have ported all provider-independent Cluster API resources to cluster chart, which was phase 1 of restructuring of cluster- apps, see https://github.com/giantswarm/roadmap/issues/2742 for more details. Now we want to use cluster chart in cluster-aws and remove all provider-independent Cluster API resources from cluster-aws.

In order to do so, first we have to refactor cluster-aws Helm values, so that cluster chart can read provider-independent values it needs. For that, we have to move current top-level properties to be under Values.global.

This pull request moves `Values.metadata` to `Values.global.metadata`.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### **Breaking change**
+
+- Move Helm values property `.Values.metadata` to `.Values.global.metadata`.
+
 ## [0.47.0] - 2023-11-07
 
 ### Added
--- a/helm/cluster-aws/ci/ci-values.yaml
+++ b/helm/cluster-aws/ci/ci-values.yaml
@@ -1,5 +1,6 @@
-metadata:
-  organization: "test"
+global:
+  metadata:
+    organization: "test"
 baseDomain: example.com
 
 connectivity:
--- a/helm/cluster-aws/ci/test-mc-proxy.yaml
+++ b/helm/cluster-aws/ci/test-mc-proxy.yaml
@@ -1,7 +1,8 @@
-metadata:
-  name: test-mc-proxy
-  organization: test
-  servicePriority: lowest
+global:
+  metadata:
+    name: test-mc-proxy
+    organization: test
+    servicePriority: lowest
 baseDomain: example.com
 connectivity:
   proxy:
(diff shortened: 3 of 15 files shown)


.github/workflows/zz_generated.diff_helm_render_templates.yaml
```diff
@@ -44,12 +44,32 @@ jobs:
           echo "values=[\"$(ls -m helm/${{ github.event.repository.name }}/ci/test-*-values.yaml | tr -d ' \n' | sed 's/,/\", \"/g')\"]" >> $GITHUB_OUTPUT
     outputs:
       values: ${{ steps.get-rendering-values.outputs.values }}
+  get-old-rendering-values:
```
nprokopic, line 47: ⚠️ Passing old Helm values from main branch when generating old Helm templates from main branch.

The action is originally (IIUC mistakenly) passing new Helm values from the current branch to the old Helm chart from the main branch.

.github/workflows/zz_generated.json_schema_validation.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Run schemalint
         id: run-schemalint
-        uses: giantswarm/schemalint/actions/verify-helm-schema@v2
+        uses: giantswarm/schemalint/actions/verify-helm-schema@support-global-property
```
nprokopic, line 32: ⚠️ This PR is temporarily using modified schema linting from https://github.com/giantswarm/schemalint/pull/120. This will be reverted when new schema linter released.

fiunchinho (approved):

.github/workflows/zz_generated.json_schema_validation.yaml
nprokopic: Reverted as we schemalint is updated.

AndiDog (approved):

.github/workflows/zz_generated.diff_helm_render_templates.yaml
nprokopic: Fixed in devctl here https://github.com/giantswarm/devctl/pull/586, so when the workflows are re-generated the changes here will not be lost.

PR #422, opened 2023-11-17
Enable prod cluster teleport by default

kopiczko:


nprokopic:
Should this PR also add the change that the "regular" bastion is disabled by default?

PR #425, opened 2023-11-20
Re-generated docs.

weseven:
### Checklist

### Trigger e2e tests

/run cluster-test-suites


calvix:
`JSON schema validation / Verify values.schema.json with schemalint (push) ` still fails is that expected?

nprokopic:
> `JSON schema validation / Verify values.schema.json with schemalint (push) ` still fails is that expected?

This is the error:

```
Either global or root-level property must be present (both are not allowed)
  - /metadata
```

Which happens because there is both `.Values.global.metadata` (new schema) and `.Values.metadata` old schema, so the lint fails, because it does not let to mix old and new schemas (the Helm values schema should be either migrated or not).

That error will go away when I merge this PR https://github.com/giantswarm/cluster-aws/pull/413 (and all other related PRs), which should happen later today.

fiunchinho (approved):

PR #427, opened 2023-11-21
rework kubelet configuration to have a file instead of bash script

calvix:
rework kubelet config from bash script into a regular file we can use for other fixes
it will be useful later for https://github.com/giantswarm/giantswarm/issues/29006

before: running bash script which uses `sed` to  append lines in kubelet config file and restart kubelet afterward, 

now: append the kubelet config struct defined here https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#kubelet to `/etc/kubeadm.yaml` file which is read by kubeadm and kubelet config is generated based on the config


helm/cluster-aws/values.schema.json
```diff
@@ -88,6 +88,11 @@
                     "type": "integer",
                     "title": "Minimum number of nodes"
                 },
+                "podShutdownGracePeriodSec": {
+                    "type": "integer",
+                    "title": "Pod shutdown grace period (seconds)",
+                    "description": "How long will kubelet wait for pods to gracefully shutdown before forcefully terminating them. Value must in seconds."
+                },
```
mnitchev: I think we can add the default here instead?

helm/cluster-aws/files/etc/kubelet-configuration.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration
+shutdownGracePeriod: 300s
+shutdownGracePeriodCriticalPods: 60s
```
fiunchinho, line 4: You are missing a new line at the end of the file.

fiunchinho (approved):

calvix:
> If you happen to change Helm values here, please ping me, as we have to port this change to cluster chart (I have already ported the "old way" with the script).

should be without helm changes

helm/cluster-aws/files/etc/kubelet-configuration.yaml
calvix: fixed

PR #430, opened 2023-11-21
Move Values.providerSpecific to Values.global.providerSpecific

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/2954

:warning: This is a Helm values breaking change, so it will require adjusting Helm values for all cluster when doing the upgrade.

We have ported all provider-independent Cluster API resources to `cluster` chart, which was phase 1 of restructuring of cluster-<provider> apps, see https://github.com/giantswarm/roadmap/issues/2742 for more details. Now we want to use `cluster` chart in `cluster-aws` and remove all provider-independent Cluster API resources from `cluster-aws`.

In order to do so, first we have to refactor `cluster-aws` Helm values, so that `cluster` chart can read provider-independent values it needs. For that, we have to move current top-level properties to be under `Values.global`.

This pull request moves `Values.providerSpecific` to `Values.global.providerSpecific`. While this property is not provider-independent, it has to be visible from the Helm context of `cluster` chart, as the `cluster` chart will need to render machine templates from `cluster-$provider` in order to correctly calculate hash used for resource name suffixes.

### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move Helm values property `.Values.nodePools` to `.Values.global.nodePools`.
 - Move Helm values property `.Values.managementCluster` to `.Values.global.managementCluster`.
 - Move Helm values property `.Values.baseDomain` to `.Values.global.connectivity.baseDomain`.
+- Move Helm values property `.Values.providerSpecific` to `.Values.global.providerSpecific`.
 
 ### Added
 
--- a/helm/cluster-aws/README.md
+++ b/helm/cluster-aws/README.md
@@ -13,16 +13,16 @@ schemadocs generate helm/cluster-aws/values.schema.json -o helm/cluster-aws/READ
 <!-- DOCS_START -->
 
 ### AWS settings
-Properties within the `.providerSpecific` top-level object
+Properties within the `.global.providerSpecific` object
 
 | **Property** | **Description** | **More Details** |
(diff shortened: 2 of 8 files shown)


calvix:
just wondering what is the use-case of provider-specific settings in the provider-independent chart `cluster`?


I thought we needed to add values under `.global` in order to be used in both charts  and this feels like something that should be in provider chart only

calvix:
Still, it feels weird to do it maybe we should ignore hash calculation on provider-specific  values, I can't remember a value where it would be needed

nprokopic:
> Still, it feels weird to do it maybe we should ignore hash calculation on provider-specific values, I can't remember a value where it would be needed

Yeah, it does feel weird.

Bastion is using `.Values.providerSpecific.flatcarAwsAccount` here https://github.com/giantswarm/cluster-aws/blob/88b82a241c9dfc5d78b32143bc93dcb888b73ba6/helm/cluster-aws/templates/_bastion.tpl#L11. That value should probably never change, so it could be possible to somehow exclude it from hash calculation.

Even without the above, I think that moving `Values.providerSpecific` to  `Values.global.providerSpecific` at this moment is a good idea ("at this moment" since we're breaking Helm values anyway), since with this one we get almost clean split between `global` and `internal` in the root of Helm values \*, which is basically a split between "public" values, that customers can override and can also be used in subcharts if needed, and "private" Helm values that should be (for some reason) internal to a specific chart.

fiunchinho (approved):

AndiDog (approved):

PR #450, opened 2023-12-11
fix-subnet-filter-for-az

calvix:
towards https://github.com/giantswarm/giantswarm/issues/29201

/run cluster-test-suites


calvix:
the fix seems to be easy but we might need to manually  fix MCs before updating to this version as they might run instances in the wrong subnets

calvix:
ASG automatically balances instances and removes the ones from the wrong AZs, which is neat.

fiunchinho (approved):

PR #455, opened 2023-12-13
remove-bastion

calvix:
towards https://github.com/giantswarm/giantswarm/issues/29254

/run cluster-test-suites


calvix:
the test fails because of test for bastion so it needs to be first disabled for CAPA - https://github.com/giantswarm/cluster-test-suites/pull/164

fiunchinho (changes requested):
Please update the changelog.

fiunchinho (approved):

PR #457, opened 2023-12-18
Change `KubeadmConfig` bootstrap config reference to ensure nodes get rolled when making changes to node specification, add instance warmup setting

AndiDog:

https://github.com/giantswarm/roadmap/issues/2217

Together with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4619 and related CAPI fixes, we can now use this new feature. By changing the bootstrap config reference, i.e. the `KubeadmConfig` name in our case, the `AWSMachinePool` reconciler will roll out new nodes via ASG instance refresh. Since that would normally replace EC2 instances as quick as possible, and since there's no health check implemented yet, I'm adding the instance warmup setting (5 minutes, not configurable for now) such that old worker nodes don't all get replaced/terminated too quickly.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


AndiDog:
Test failures are unrelated ([chat](https://gigantic.slack.com/archives/C0251MXL5/p1702987615760869)), but I still want them go through.

/run cluster-test-suites

helm/cluster-aws/templates/_machine_pools.tpl
```diff
@@ -93,6 +169,8 @@ spec:
     {{- end }}
     {{- end }}
   {{- end }}
+  refreshPreferences:
+    instanceWarmup: 300
```
fiunchinho, line 173: Maybe worth adding a note in our docs about this behavior.

fiunchinho (approved):

helm/cluster-aws/templates/_machine_pools.tpl
AndiDog: @fiunchinho where would be a good place apart from the changelog?

Also mind that cluster-autoscaler changes to the ASG won't have this setting. It's only for CAPA-triggered instance refresh and doesn't apply to the ASG in general. We'd have to implement the `CreateAutoScalingGroup: DefaultInstanceWarmup` parameter in CAPA. Until then, my gut says this change is quite an implementation detail which could still change and probably needs no special note. But I'm also okay to somehow add it to https://docs.giantswarm.io/advanced/cluster-management/node-pools-capi/ if you think it's reasonable. I'm creating a follow-up issue for the general instance warmup parameter.

fiunchinho: I'd still mention it , so everyone has the full picture on what to expect when we perform an upgrade. [That's the section I was thinking](https://docs.giantswarm.io/advanced/cluster-management/node-pools-capi/#what-happens-when-rolling-nodes).

PR #472, opened 2024-01-09
Use 443 as default api-server Load Balancer port

mnitchev:
### Checklist

### Trigger e2e tests

/run cluster-test-suites

### DIFF
--- a/.circleci/ct-config.yaml
+++ b/.circleci/ct-config.yaml
@@ -1,4 +1,4 @@
 ---
 chart-repos:
   - cluster-catalog=https://giantswarm.github.io/cluster-catalog
-  - cluster-catalog=https://giantswarm.github.io/cluster-test-catalog
+  - cluster-test-catalog=https://giantswarm.github.io/cluster-test-catalog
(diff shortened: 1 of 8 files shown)


fiunchinho (commented):
What about calling it `apiServerLBPort` instead? I remember having trouble differentiating the two while investigating this issue.

mnitchev:
Already had this discussion here https://github.com/giantswarm/cluster/pull/42#discussion_r1453599637
The value is already in the cluster chart

fiunchinho:
I get that it's confusing within CAPI and CAPI providers (hence my first comment), but why keep using a potentially confusing field name when we could be more specific in our values?

mnitchev:
Setting it to `apiServerLBPort` won't help with confusion IMO though. It makes sense in CAPA and in CAPZ with caveats, but in CAPI it has a different meaning and in other providers it might have none or something completely different. I guess the ideal thing here would be to have the value in the cluster chart be internal and named exactly as the thing that's being set as to remove ambiguity. So something like
```yaml
internal:
  clusterNetwork:
    apiServerPort: 6443
```

And then we have a separate value here that's named like the thing we _actually_ want to set and use a yaml anchor to configure the internal value like this:
```yaml
global:
  controlPlane:
    apiServerLoadBalancerPort: &lbPort 443
    ...
cluster:
  internal:
    clusterNetwork:
    apiServerPort: *lbPort
```
Not sure if that would work exactly though. But WDYT of the idea?

fiunchinho (approved):

PR #480, opened 2024-01-16
Use squared icon

fiunchinho:
Use squared icon.

### Checklist


whites11 (approved):
that seems a very good and important change we should merge ASAP

PR #507, opened 2024-02-07
Add network-policies helm release

mnitchev:
### Checklist

### Trigger e2e tests

/run cluster-test-suites


nprokopic:
Do we need these on all providers? (I guess yes.) In that case I think we can add them to the cluster chart right away.

mnitchev:
@nprokopic I think this has already been done on CAPV and CAPVCD and besides CAPA here only CAPZ doesn't have this changed, but I'm not sure what the state there is and if it's ready for it as there were some cilium changes that needed to be done first.

nprokopic:
> @nprokopic I think this has already been done on CAPV and CAPVCD and besides CAPA here only CAPZ doesn't have this changed, but I'm not sure what the state there is and if it's ready for it as there were some cilium changes that needed to be done first.

Thanks for the info. This sounds then that we really need it for all providers, meaning that we should have this in the cluster chart. I mean, you can add it here now, but then we'll just probably move it to the cluster chart (probably in the following weeks) together with cilium and all other provider-independent HelmReleases that are currently in cluster-aws.

Provider-independent `.global.*` values for features like this (like `global.connectivity.network.allowAllEgress` here) should be something that makes sense and works for all providers (this here looks like that). And in general, if there are any differences across different providers, then the resources would be just more configurable via Helm values (this is probably not the case here for network policies, as this here looks like it would be identical for all providers).

P.S. I just wanted to comment here, so we avail double work if possible in any way.

mnitchev:
@nprokopic I agree this should end up in the cluster chart, but I would rather do it separately and not block this PR on it. I think we first need to spend the time to bring this change to CAPZ and then spend the effort to extract it into the cluster chart when all providers are ready.

calvix (approved):

PR #523, opened 2024-02-23
Use cilium and network-policies from cluster chart

nprokopic:
Closes https://github.com/giantswarm/roadmap/issues/2851

Use cilium and network-policies from cluster chart, and remove them from cluster-aws.

### Checklist

### Trigger e2e tests

/run cluster-test-suites


helm/cluster-aws/templates/_cilium_helmrelease_config.yaml
```diff
@@ -0,0 +1,33 @@
+{{/* AWS-specific cilium Helm values*/}}
```
nprokopic, line 1: Here we now have only AWS-specific cilium config values.

Provider-independent default config is in cluster chart (I have compared cilium config across providers and put the common config to cluster chart). AWS-specific config can add new and/or overwrite existing provider-independent config values.

njuettner (approved):

PR #526, opened 2024-02-27
Remove leftovers which moved to cluster chart

AndiDog:

Title says it all. I checked that the rendered result before and after this change are identical.

### Checklist

- [ ] Update changelog in CHANGELOG.md. => No because there's no customer-visible change.

### Trigger e2e tests

/run cluster-test-suites


weseven (approved):

calvix (approved):

calvix:
Its  wird that the output is identical as I would expect removed duplicity of the `systemd-unit`

calvix:
> > Its weird that the output was identical as I would expect removed duplicity of the `systemd-unit` for `os-hardening`
> > is `flatcarSystemdUnits` function even used anywhere?
> 
> @calvix Good catch. And there was a lot more to remove. I updated the PR. Can you check if it was on purpose to make `internal.migration.apiBindPort` meaningless? I've removed it but maybe we need to bring it back as `global.[...]`. Same for `internal.cgroupsv1`.
> 
> /run cluster-test-suites

I think most of the values moved to different place since we are using `cluster` chart 

there should be a PR that fixes it for the migration CLI
https://github.com/giantswarm/capi-migration-cli/pull/15/files

So should be fine

calvix (approved):

PR #556, opened 2024-03-21
Smart availability zones

whites11:

Towards: https://github.com/giantswarm/giantswarm/issues/30316

This PR makes 3 changes

- changes the helm chart to use a yaml file (in files/azs-in-region.yaml) to get a list of azs in a region when defaulting such value
- adds a go tool in the `azs-getter` folder to generate the above file automatically
- adds a github action that runs once a day the above tool and creates a PR if any change happens

this is to replace the previous logic that hardcoded "a", "b" and "c" as az name for each region

### Checklist

### Trigger e2e tests

/run cluster-test-suites


helm/cluster-aws/templates/_awsregion.tpl
```diff
@@ -12,5 +12,5 @@ If no region is provided in the values we'll attempt to look it up based on the
 {{- end }}
 {{- end }}
 {{- end }}
-{{- $region }}
+{{- required "Unable to default the region. Please set .global.providerSpecific.region" $region }}
```
whites11, line 15: A bit out of scope for this PR.
We provide the "autodetect region" feature as an helper, but we should error if we're unable to provide a sane default.

calvix (approved):
LGTM

whites11:
skipping CI blocker. failures are in the deletion phase and being worked on.

PR #572, opened 2024-04-04
Update Availability Zones in helm/cluster-aws/files/azs-in-region.yaml

taylorbot:


Gacko (approved):

helm/cluster-aws/files/azs-in-region.yaml
```diff
@@ -72,7 +76,7 @@ us-east-2:
     - c
 us-west-1:
     - a
-    - b
+    - c
```
Gacko, line 79: @whites11 Is that intended?

whites11: it's known let's say.
us-west-1 only has 2 AZs and they change name depending which account you connect the AWS API with ... 
We decided to live with this

Gacko: That sucks... I enabled auto-merge for this PR. Hope that's fine for you! Just want to get rid of the piled up automation PRs. :)

PR #582, opened 2024-04-11
Repository: Chores.

Gacko:
### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


Makefile.custom.mk
```diff
@@ -1,14 +0,0 @@
...
-
-ensure-schema-gen:
-	@helm schema-gen --help &>/dev/null || helm plugin install https://github.com/mihaisee/helm-schema-gen.git
-
-.PHONY: schema-gen
-schema-gen: ensure-schema-gen ## Generates the values schema file
-	@cd helm/cluster-aws && helm schema-gen values.yaml > values.schema.json
-
-.PHONY: update-chart-deps
-update-chart-deps: ## Update chart dependencies to latest (matching) version
```
nprokopic, line 13: This one is useful. Not sure about other people, but I use it while developing (yes, so I don't have to `cd` in and out of `helm` dir 😬).

Gacko: This is already included in `Makefile.gen.app.mk`:

```sh
APPLICATION=helm/cluster-aws make update-deps
```

I agree, it's not that straight forward, but well. Not sure if we should maintain the same code twice.

nprokopic: didn't see that one 👍

Gacko:
Flaky little ba****d 😈

/run cluster-test-suites

PR #600, opened 2024-04-25
Add additional EBS volume to control-plane nodes for `/var/log`

calvix:
towards https://github.com/giantswarm/giantswarm/issues/29019

- Add `log` volume to control-plane nodes.

default size 15GB equals to value that we use for vintage as a default


helm/cluster-aws/values.schema.json
```diff
@@ -966,6 +995,11 @@
                                 "title": "Address range"
                             }
                         },
+                        "logVolumeSizeGB": {
+                            "type": "integer",
+                            "title": "Log volume size (GB)",
+                            "default": 15
```
whites11, line 1001: 15 sounds like a very small default to me

calvix: thats what we do in vintage

calvix: So can you give me a better default? I am all ears I just don't have an idea

PR #610, opened 2024-05-06
Testing SELinux in enforcing mode

weseven:

Testing enabling selinux in enforcing mode from cluster chart: https://github.com/giantswarm/cluster/pull/170

Towards https://github.com/giantswarm/giantswarm/issues/30605


weseven:
Standard & upgrade e2e tests are fine, the ones in the private cluster hit the "app stuck in pending-install/pending-upgrade" issue so they will probably fail, but it's not an SELinux dependent issue.

I've tested the creation of a cluster with SELinux in enforcing mode on golem successfully (simply setting `global.components.selinux.mode: enforcing`  in the cluster userconfig configmap).

PR #612, opened 2024-05-07
Add selinux config

weseven:

Towards https://github.com/giantswarm/giantswarm/issues/30605

This PR updates to cluster 0.22.0 and allows to set SELinux mode (enforcing/permissive/disabled) through helm value `global.components.selinux.mode`.

Default mode is permissive, same as current flatcar version.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


weseven:
Standard & upgrade tests are fine, private fails for the app stuck in pending-upgrade issue (unrelated to the PR, which in any case keeps the flatcar default of permissive).

njuettner (approved):

nprokopic (approved):

PR #618, opened 2024-05-09
Release v0.75.0

taylorbot:


uvegla:
Nice! This is exactly what I need to test: https://github.com/giantswarm/giantswarm/issues/30596 (`giantswarm/cluster` release `v0.20.0` or newer).

Just one note I bumped into with the previous `0.74.0` (maybe this fixes that): in Happa because of the previous schema change the default validation fails, for example: https://happa.golem.gaws.gigantic.io/organizations/giantswarm/clusters/new. Not present in `0.73.0`.

weseven:
> Nice! This is exactly what I need to test: [giantswarm/giantswarm#30596](https://github.com/giantswarm/giantswarm/issues/30596) (`giantswarm/cluster` release `v0.20.0` or newer).
> 
> Just one note I bumped into with the previous `0.74.0` (maybe this fixes that): in Happa because of the previous schema change the default validation fails, for example: https://happa.golem.gaws.gigantic.io/organizations/giantswarm/clusters/new. Not present in `0.73.0`.

I don't think this will fix that, for some strange reason Happa is complaining that those tags should be an array while in the schema they are actually defined as a map/object: https://github.com/giantswarm/cluster-aws/blob/1cb397b50c815f02cc92e88024226e5e6e88805d/helm/cluster-aws/values.schema.json#L719
They also all have default values which Happa is not rendering, so no clue what's wrong: maybe on Happa's side the library doing the schema validation does not like some references? 
@AndiDog have you got any ideas?

weseven:
> Hold on, this might be breaking containerd: https://github.com/giantswarm/cluster/blob/v0.22.0/helm/cluster/files/etc/containerd/config.toml#L29 based on: https://gigantic.slack.com/archives/C0559SH3RJ4/p1715178031451949?thread_ts=1715158412.440559&cid=C0559SH3RJ4
> 
> cc. @nprokopic can confirm / disprove that better I think.
> 
> Because of this:
> 
> https://github.com/giantswarm/cluster-aws/blob/cc9653ca62b2f8987fcd43a770f7654cde216dd6/helm/cluster-aws/Chart.yaml#L18-L19

You're right, it doesn't actually break (the clusters are created without issues), but anyway we should fix the configs in cluster :)

weseven:
Closing this PR, will do a new release once we update the cluster chart.

PR #627, opened 2024-05-17
Replace IPV4_LOCAL with COREOS_IPV4_LOCAL value from coreos-metadata service

njuettner:

In order to support all providers we replaced existing `COREOS_EC2_IPV4_LOCAL` and `COREOS_EC2_HOSTNAME` in `cluster` chart and add a support to use environment variables.

Nothing will change for CAPA and it will work the same as before.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


calvix (approved):
LGTM, just to confirm my understanding, each chart has customer values for the ENV and we set it via the providerIntegration values?

Gacko (approved):
LGTM, do not forget to reset the `Chart.yaml` & `Chart.lock`. 😉

njuettner:
> LGTM, just to confirm my understanding, each chart has customer values for the ENV and we set it via the providerIntegration values?

Correct

njuettner:
Tests were green except `Cilium ENI mode tests scale [It] scales node by creating anti-affinity pods` it was due to spot instances not being available.

We have the same test for CAPA Standard and this was working fine. I'll merge this PR


https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-aws-627-cluster-test-suitesmmh5w?pipelineTask=run-tests&taskRunName=pr-cluster-aws-627-cluster-test-suitesmmh5w-run-tests-3&step=run-tests

PR #639, opened 2024-06-05
Fix AWS pod identity webhook app userValues to be always rendered

calvix:
the configmap for the app si always rendered but the user config was not, this broke automatic detection for china values and broke the functionality


Gacko (approved):
Nice catch. I implemented it with optional user values back then, but I think Nikola finished my work and added the additional values without removing the condition in the `userConfig` section.

calvix:
the error only happened in `upgrade` pipeline and seems to be unrelated as the failed app was kyverno
```
  {"level":"info","ts":"2024-06-05T12:30:45Z","msg":"App status for 't-q6udkbwn6t9hxa71al-kyverno' is not yet as expected: expectedStatus='deployed' actualStatus='pending-upgrade' (reason: 'Preparing upgrade')"}
  ```

PR #647, opened 2024-06-06
Set maximum number of pods (kubelet `--max-pods`) for Cilium ENI mode due to restrictions by number of ENIs and IPs per ENI

AndiDog:

Towards https://github.com/giantswarm/roadmap/issues/2563

This comes after the required user data size limit avoidance in https://github.com/giantswarm/cluster-aws/pull/641, therefore stacking _this_ PR on top.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

Later, when the dependent PR was merged.


helm/cluster-aws/files/opt/bin/kubelet-aws-config.sh
```diff
@@ -0,0 +1,547 @@
...
+*)
+	max_pods=40
+	echo "WARNING: EC2 instance type '${instance_type}' is not covered in ${0}, defaulting to max_pods=${max_pods}"
+	;;
+esac
+
+# We don't want to have more than 110 pods on a node even it would be possible by IPs
+if (($max_pods > 110)); then
+	max_pods=110
+fi
```
fiunchinho, line 537: Do we know why? I'd document the reasoning here too.

fiunchinho (approved):

fiunchinho:
Only the upgrade tests failed and it seems to be a known issue, so merging anyway.

helm/cluster-aws/files/opt/bin/kubelet-aws-config.sh
AndiDog: I unfortunately don't know. This was taken over from before.

fiunchinho: [I found this](https://kubernetes.io/docs/setup/best-practices/cluster-large/). I think this must be the reason.

PR #657, opened 2024-06-13
Bumping cluster app with necessary changes

ljakimczuk:
### What this PR does

This PR bumps Cluster Chart to version 0.33.1 which introduces the [Giantswarm Cluster Suite app](https://github.com/giantswarm/giantswarm-cluster-suite). The GCS app is disabled by default, hence must be enabled in this app.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add the Management Cluster name as a tag to the AWS resources created by CAPA.
 - Add the node pool name as a tag to the AWS resources associated with the node pool.
 
+### Changed
+
+- Update cluster chart to 0.33.1
+
 ## [1.0.0] - 2024-06-20
 
 ### Changed
 
 - First major release, breaking changes not allowed in minor releases anymore.
-- Update cluster chart to 0.33.1
 
 ## [0.79.0] - 2024-06-19
 


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - First major release, breaking changes not allowed in minor releases anymore.
+- Update cluster chart to 0.33.1
```
fiunchinho, line 20: It looks like a merge screw up the changelog. This is the wrong block.

ljakimczuk: Ah, I guess I am really bad at modifying changelog 😭 Let me fix that

fiunchinho:
Regarding `giantswarm-cluster-suite`, for simple resources like a `Namespace` and a `PriorityClass`, wouldn't it be simpler to inline them and apply them as [a CAPI `ClusterResourceSet`](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-resource-set) defined in [the `cluster` chart](https://github.com/giantswarm/cluster)? Do you think that approach would be more complicated? I think it's simpler. and easier to reason about. 

There wouldn't be a reason to maintain yet another `App` and a Github repository for that. It wouldn't be necessary to take care of dependencies between apps while creating a cluster. And updating these resources (I mean the `Namespace` and `PriorityClass`) would be only one PR to the `cluster` chart, instead of two of them (one on the app repo and then a version bump in the `cluster` chart). Finally, creating an `app` as a all-in-one bag having a mixture of different resources with no connection between them feels wrong somehow.

This feels like a more CAPI native way of solving our requirements. A `ClusterResourceSet` is a CAPI object that applies to the WC some manifests defined in a `ConfigMap` (not just once, it will reconcile/re-apply the contents if the contents of the `ConfigMap` change). You can find more details in [the design proposal](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200220-cluster-resource-set.md).
In this case, the `Namespace` and the `PriorityClass` would be defined in the `cluster` chart, and whenever we need to update those (if ever) or new resources are needed, they would get re-applied by CAPI.

What do you think?

fiunchinho:
Thank you.

ljakimczuk:
@fiunchinho I discussed that within the team and to be honest, we are indifferent to how these two resources are provided to the cluster, this seems however a bit troublesome to change the approach now, for this app is already part of Cluster app and we initiated changes to Release CRs as well. On top of that, creation of this new app has been agreed a time ago, see this [Slack thread](https://gigantic.slack.com/archives/C01F7T2MNRL/p1718099832629989), so we would prefer continuing this approach, just to get the changes we need live as soon as possible, but then at some point if you have better solution we could replace it with that.

ljakimczuk:
Not sure though why test fails for China. The first run reported:

```text
Summarizing 1 Failure:
  [FAIL] Common tests hello world [It] should have a ready Certificate generated
  /app/internal/common/hello.go:243
```

while the second:

```text
Summarizing 3 Failures:
  [FAIL] Common tests basic [It] has all its Deployments Ready (means all replicas are running)
  /app/internal/common/basic.go:82
  [FAIL] Common tests hello world [It] cluster wildcard ingress DNS must be resolvable
  /app/internal/common/hello.go:121
  [FAIL] Common tests scale [It] scales node by creating anti-affinity pods
  /app/internal/common/scale.go:146
```

What seems like unrelated to my changes, maybe.

ljakimczuk:
/run cluster-test-suites TARGET_SUITES=./providers/capa/china

Gacko (changes requested):
Independent of where this discussion is ending up: Could you please update your PR to `cluster` v0.34.1 as soon as it has been released? I approved your PR. That would be awesome! Thank you!

PR #699, opened 2024-07-18
Feature Gates: Remove `CronJobTimeZone`.

Gacko:

This PR removes the `CronJobTimeZone` feature gate as it becomes stable and is included in Kubernetes v1.29.

For Kubernetes <v1.29, you will need to re-enable it using the respective values.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


Gacko:
Note: This PR is going to be tested from the CAPA v29.0.0 PR: https://github.com/giantswarm/releases/pull/1318

I therefore put the test version of the chart in the release as the normal cluster test suite cannot rely on a yet unreleased CAPA version.

fiunchinho (approved):

Gacko:
Cluster Test Suites fail due to an installation timeout. I already proposed [a PR](https://github.com/giantswarm/cluster-aws/pull/708) for the service monitor apps, but since AWS Pod Identity Webhook still depends on `cert-manager`, we might need to review the dependency tree and/or improve timeouts.

PR #708, opened 2024-07-22
Apps: Fix service monitor dependencies.

Gacko:

I assume those two apps have been copied and pasted from AWS Pod Identity Webhook app and therefore still contained the `cert-manager` dependency. Actually they should depend on the `observability-bundle`.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Apps: Fix service monitor dependencies.
+
 ## [1.2.0] - 2024-07-22
 
 This release removes the `CronJobTimeZone` feature gate as it becomes stable and is included in Kubernetes v1.29.
--- a/helm/cluster-aws/templates/aws-ebs-csi-driver-servicemonitors-app.yaml
+++ b/helm/cluster-aws/templates/aws-ebs-csi-driver-servicemonitors-app.yaml
@@ -15,7 +15,7 @@ metadata:
     {{- include "labels.common" . | nindent 4 }}
     giantswarm.io/managed-by: {{ .Chart.Name }}
   annotations:
-    app-operator.giantswarm.io/depends-on: {{ include "resource.default.name" . }}-cert-manager
+    app-operator.giantswarm.io/depends-on: {{ include "resource.default.name" . }}-prometheus-operator-crd
 spec:
   catalog: default
   name: aws-ebs-csi-driver-servicemonitors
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ For Kubernetes <v1.29, you will need to re-enable it using the respective values
 ### Changed
 
 - Chart: Update `cluster` chart to v0.36.0. ([#703](https://github.com/giantswarm/cluster-aws/pull/703))
+- Apps: Fix service monitor dependencies. ([#708](https://github.com/giantswarm/cluster-aws/pull/708))
```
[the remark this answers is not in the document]
Gacko: So people can get detailed information more easily. Of course changelogs should already contain all the details, but I did this in other apps in the past and it helped a lot from time to time, especially for customers.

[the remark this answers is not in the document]
fiunchinho: I'm fine discussing that approach, it could be an improvement. But let's agree on the format before each of us follow a different format. Until then, I'd drop the link.

helm/cluster-aws/templates/aws-ebs-csi-driver-servicemonitors-app.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
     {{- include "labels.common" . | nindent 4 }}
     giantswarm.io/managed-by: {{ .Chart.Name }}
   annotations:
-    app-operator.giantswarm.io/depends-on: {{ include "resource.default.name" . }}-cert-manager
+    app-operator.giantswarm.io/depends-on: {{ include "resource.default.name" . }}-observability-bundle
```
AverageMarcus, line 18: Shouldn't this be `prometheus-operator-crd` ?

Example from old default-apps-aws - https://github.com/giantswarm/default-apps-aws/blob/e20e94cced3369d690f0915134cbd8e5fac49b2c/helm/default-apps-aws/values.yaml#L159

AverageMarcus: And from cilium-servicemonitors from cluster - https://github.com/giantswarm/cluster/blob/6017f6ed54d4152df6df94d5dbf04dae2c72e502/helm/cluster/files/apps/cilium-servicemonitors.yaml#L7

Gacko: You're right! Actually the little chat we had on Slack also reminded me of the fact that there's a separate app for those CRDs. Gonna update my PR, thank you!

Gacko:
I implemented your suggestions.

AverageMarcus (approved):

Gacko:
I' quite a bit surprised: All tests pass in the first run after my last change. 😍

PR #780, opened 2024-08-22
Validate that machine pool availability zones belong to the selected region

fiunchinho:

Sometimes when configuring node pools, users may forget to choose the right availability zones, and select AZs that differ from the selected region. This makes cluster creation to fail, but the user needs to investigate why the cluster is not coming up. Instead, it would be better to fail earlier, explaining the user what the error is.

### Checklist

- [X] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites

### DIFF
--- a/helm/cluster-aws/templates/_awsazs.tpl
+++ b/helm/cluster-aws/templates/_awsazs.tpl
@@ -2,10 +2,20 @@
 If no availability zones are provided in the values we'll attempt to look it up based on the region specified in AWSCluster if its missing as well, AZs of the management cluster are used
 */}}
 {{- define "aws-availability-zones" }}
+{{- $region := include "aws-region" . }}
 {{- if .mp.availabilityZones }}
 {{- .mp.availabilityZones | toYaml }}
+{{- $isValid := true -}}
+{{- range $az := .mp.availabilityZones -}}
+  {{- $azRegion := (printf "%s" $az | trunc 9) -}}
+  {{- if ne $azRegion $region -}}
+    {{- $isValid = false -}}
+  {{- end -}}
+{{- end -}}
+{{- if not $isValid -}}
+  {{- fail (printf "Invalid MachinePool Availability Zones provided for region %s. Availability Zones: %v" $region .mp.availabilityZones) -}}
+{{- end -}}
 {{- else }}
-{{- $region := include "aws-region" . }}
 {{- include "azs-in-region" (dict "region" $region  "Files" .Files ) }}
 {{- end }}
 {{- end }}
--- a/helm/cluster-aws/templates/_awsazs.tpl
+++ b/helm/cluster-aws/templates/_awsazs.tpl
@@ -5,16 +5,11 @@ If no availability zones are provided in the values we'll attempt to look it up 
 {{- $region := include "aws-region" . | trim }}
 {{- if .mp.availabilityZones }}
 {{- .mp.availabilityZones | toYaml }}
-{{- $isValid := true -}}
-{{- range $az := .mp.availabilityZones -}}
-  {{- $azRegion := (printf "%s" $az | trunc 9) -}}
-  {{- if ne $azRegion $region -}}
-    {{- $isValid = false -}}
-  {{- end -}}
-{{- end -}}
-{{- if not $isValid -}}
-  {{- fail (printf "Invalid value in global.nodePools.*.availabilityZones: the specified availability zones %v are not in the cluster region %q" .mp.availabilityZones $region) -}}
-{{- end -}}
+{{- range $az := .mp.availabilityZones }}
+  {{- if not (hasPrefix $region $az) }}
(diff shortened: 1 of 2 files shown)


helm/cluster-aws/templates/_awsazs.tpl
```diff
@@ -2,10 +2,20 @@
 If no availability zones are provided in the values we'll attempt to look it up based on the region specified in AWSCluster if its missing as well, AZs of the management cluster are used
 */}}
 {{- define "aws-availability-zones" }}
+{{- $region := include "aws-region" . }}
```
AndiDog, line 5: [minor]

Another code location in this chart is using `trim`. This would prevent us from failing here if the `aws-cluster` definition returns whitespace by mistake.

```suggestion
{{- $region := include "aws-region" . | trim }}
```

helm/cluster-aws/templates/_awsazs.tpl
```diff
@@ -2,10 +2,20 @@
...
 {{- if .mp.availabilityZones }}
 {{- .mp.availabilityZones | toYaml }}
+{{- $isValid := true -}}
+{{- range $az := .mp.availabilityZones -}}
+  {{- $azRegion := (printf "%s" $az | trunc 9) -}}
+  {{- if ne $azRegion $region -}}
+    {{- $isValid = false -}}
+  {{- end -}}
```
AndiDog, line 13: `trunc 9` is too specific. For `eu-central-1a`, this would give `eu-centra`.

Please add one or two test files `ci/test-*-values.yaml` to ensure that a set of valid values (e.g. region `eu-central-1`, machine pool `[...].availabilityZones` `eu-central-1{a,b,c}`) work fine and we cannot break the chart.

I think prefix matching should work?! =>

```suggestion
  {{- $azRegion := (printf "%s" $az | trunc 9) -}}
  {{- if not (hasPrefix (printf "%s-" $region) $az) -}}
    {{- $isValid = false -}}
  {{- end -}}
```

helm/cluster-aws/templates/_awsazs.tpl
```diff
@@ -2,10 +2,20 @@
...
+{{- end -}}
+{{- if not $isValid -}}
+  {{- fail (printf "Invalid MachinePool Availability Zones provided for region %s. Availability Zones: %v" $region .mp.availabilityZones) -}}
```
AndiDog, line 16:
```suggestion
  {{- fail (printf "Invalid value in global.nodePools.*.availabilityZones: the specified availability zones %v are not in the cluster region %q" .mp.availabilityZones $region) -}}
```

helm/cluster-aws/templates/_awsazs.tpl
```diff
@@ -2,10 +2,20 @@
...
+{{- $isValid := true -}}
+{{- range $az := .mp.availabilityZones -}}
+  {{- $azRegion := (printf "%s" $az | trunc 9) -}}
+  {{- if ne $azRegion $region -}}
+    {{- $isValid = false -}}
+  {{- end -}}
+{{- end -}}
+{{- if not $isValid -}}
+  {{- fail (printf "Invalid value in global.nodePools.*.availabilityZones: the specified availability zones %v are not in the cluster region %q" .mp.availabilityZones $region) -}}
+{{- end -}}
```
Gacko, line 17:
```suggestion
{{- $isValid := true }}
{{- range $az := .mp.availabilityZones }}
  {{- if not (hasPrefix $region $az) }}
    {{- $isValid = false }}
  {{- end }}
{{- end }}
{{- if not $isValid }}
  {{- fail (printf "Invalid value in global.nodePools.*.availabilityZones: the specified availability zones %v are not in the cluster region %q" .mp.availabilityZones $region) }}
{{- end }}
```

helm/cluster-aws/templates/_awsazs.tpl
```diff
@@ -2,10 +2,15 @@
...
 */}}
 {{- define "aws-availability-zones" }}
+{{- $region := include "aws-region" . | trim }}
 {{- if .mp.availabilityZones }}
 {{- .mp.availabilityZones | toYaml }}
+{{- range $az := .mp.availabilityZones }}
+  {{- if not (hasPrefix $region $az) }}
+    {{- fail (printf "Invalid value in `global.nodePools.*.availabilityZones`: The specified availability zone %s is not in the cluster's region %s" $az $region) }}
+  {{- end }}
+{{- end }}
```
Gacko, line 12:
```suggestion
{{- range $az := .mp.availabilityZones }}
  {{- if not (hasPrefix $region $az) }}
    {{- fail (printf "Invalid value in `global.nodePools.*.availabilityZones`: The specified availability zone %s is not in the cluster's region %s" $az $region) }}
  {{- end }}
{{- end }}
{{- .mp.availabilityZones | toYaml }}
```

Gacko (approved):

Gacko:
Good ol' China...

PR #792, opened 2024-08-28
CI: Bump release version.

Gacko:

This PR bumps the release version used in CI values to the latest available CAPA release.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


fiunchinho (approved):

fiunchinho:
This is only for helm templating, right? Does it matter which release is used?

Gacko:
Yes, it matters as soon as you're using `make template-in-cluster` to check if your changes also work with real life releases.

Normally the templates are implemented to return null, empty string or something like that in dry-run mode. But with `template-in-cluster` we tell Helm to actually use resources from the cluster you're connected to and therefore your `helm template` output can contain real life values, which is very valuable for development - at least for what I'm currently working on.

PR #794, opened 2024-08-29
Apps: Use `catalog` from Release CR.

Gacko:

This PR aligns the way we define the catalog of an app to the way we do it in the `cluster` chart and therefore allows setting a different catalog in the Release CR, which is sometimes required for testing apps during development, e.g. by setting `catalog: default-test`.

I came across this issue when forging CAPA v29.0.0, where I also needed to bump the `cloud-provider-aws` version.

As the current latest release CAPA v29.1.0 does not define the catalog for the `cilium-crossplane-resources` app, yet, this PR can not be properly tested as Cluster Test Suites pick the latest available release and patch the `cluster-aws` chart to the version provided by this PR. This issue is about to be fixed in this PR: https://github.com/giantswarm/releases/pull/1355. So please also have a look there and review it, if not already merged.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

`/run cluster-test-suites`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Do not allow additional properties in most values in order to avoid unnoticed typos.
 - Validate that machine pool availability zones belong to the selected region.
-- CI: Bump release version. ([#792](https://github.com/giantswarm/cluster-aws/pull/792))
+- CI: Bump release version.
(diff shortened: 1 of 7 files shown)


Gacko:
Changes in the rendered manifests are expected, as they do not consume a real Release CR installed in an MC. The same goes for default apps included in `cluster` chart, they also do not have a catalog or version set when rendered this way.

fiunchinho (approved):

PR #808, opened 2024-09-04
CI: Add example with subnet tags.

fiunchinho:

I was not sure about how to use the subnet tags, so I figured adding an example would help us in the future.


AndiDog (approved):

AndiDog:
Fixed the build for you, let's merge as-is

Gacko (approved):

PR #814, opened 2024-09-05
Chart: Support multiple service account issuers.

AndiDog:

Towards https://github.com/giantswarm/giantswarm/issues/31589.

Use `cluster` chart changes in https://github.com/giantswarm/cluster/pull/323 to support multiple service account issuers as a feature, and thereby get rid of `internal.migration.*`.

I'm also introducing a new, plural-named annotation for IRSA domains. We'll also need that for Karpenter, so maybe I'll still put the annotation on `AWSCluster` as well for easier consumption by `aws-crossplane-cluster-config-operator` (TODO for me, therefore still draft PR, but ready to review already).

I've created the `release-v1.1.x` branch manually, and this PR merges into it so we can create v1.1.1+ from it, with the goal of bumping to CAPA release v25.1.1+ which will be used for future migrations. And of course, we'll also merge this change into future versions of cluster-aws.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

Later, since the `cluster` chart changes must be merged first.


helm/cluster-aws/templates/_control_plane.tpl
```diff
@@ -50,14 +50,17 @@ subnet:
     {{- end }}
 {{- end }}
 
+{{- define "service-account-issuers-comma-separated" }}
+{{- range $serviceAccountIssuerIndex, $serviceAccountIssuer := .Values.cluster.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuers }}
```
iuriaranda, line 54: will this include the default issuer that's automatically composed as `https://irsa.{{ include "resource.default.name" $ }}.{{ required "global.connectivity.baseDomain value is required" .Values.global.connectivity.baseDomain }}` [here](https://github.com/giantswarm/cluster-aws/blob/9aca88c87c2df01936d11cf057687a7dcb6623ed/helm/cluster-aws/templates/_helpers.tpl#L125C1-L125C158)?

AndiDog: Yes. That Helm template function you linked is called `awsIrsaServiceAccountIssuer`, and I'm keeping the default (renamed to plural) which calls it:

```
serviceAccountIssuers:
  - templateName: awsIrsaServiceAccountIssuer
```

(quoting the generated `values.yaml`)

helm/cluster-aws/Chart.yaml
```diff
@@ -16,8 +16,8 @@ restrictions:
     - capa
 dependencies:
   - name: cluster
-    version: "0.35.0"
-    repository: https://giantswarm.github.io/cluster-catalog
+    version: "1.2.1-5995e24e7a28a3a8e598e1dab9787589b5370a84"  # actually based on `0.35.0` but CI takes the latest tag (TODO change to v0.35.1+ once `cluster` chart is released)
```
fiunchinho, line 19: don't forget about this :slightly_smiling_face:

fiunchinho (approved):
LGTM

PR #828, opened 2024-09-13
bump cluster chart to 0.35.1

mnitchev:
### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


AndiDog:
> Let's not bump the `cluster` chart twice and rather include this in @AndiDog PR.

I'd rather put changes that belong together in separate PRs. As long as we're releasing cluster-aws together, is there an issue?

Gacko:
Well, I assumed this change to be already tested if there has been a release of the `cluster` chart. So why not just bump the `cluster` chart in the `cluster-aws` only once? Your tests of the `cluster` chart already include the v0.35.1 version, right?

PR #841, opened 2024-09-18
Chart: Support multiple service account issuers.

Gacko:


helm/cluster-aws/Chart.yaml
```diff
@@ -16,8 +16,8 @@ restrictions:
     - capa
 dependencies:
   - name: cluster
-    version: "1.0.0"
-    repository: https://giantswarm.github.io/cluster-catalog
+    version: "0.35.2-49f53d3c05540c722375a848c902feddb1b614ac"
```
AndiDog, line 19: Depends on https://github.com/giantswarm/cluster/pull/339 – I approved that one as well

AndiDog (approved):

PR #912, opened 2024-10-16
Expose maxHealthyPercentage field

fiunchinho:

Towards https://github.com/giantswarm/giantswarm/issues/31305

### Checklist

- [X] Updated CHANGELOG.md.


helm/cluster-aws/values.schema.json
```diff
@@ -196,13 +196,24 @@
                     "description": "Size of the volume mounted at `/var/log` on the worker nodes.",
                     "default": 30
                 },
+                "maxHealthyPercentage": {
+                    "type": "integer",
+                    "title": "Minimum percentage of instances that must remain healthy during node replacement.",
+                    "description": "The percentage of capacity in ASG that can be in service and healthy, or pending, to support your workload when replacing instances. A larger range increases the number of instances that can be replaced at the same time. The difference between minHealthyPercentage and maxnHealthyPercentage cannot be greater than 100.",
```
Gacko, line 202:
```suggestion
                    "description": "The percentage of capacity in ASG that can be in service and healthy, or pending, to support your workload when replacing instances. A larger range increases the number of instances that can be replaced at the same time. The difference between minHealthyPercentage and maxHealthyPercentage cannot be greater than 100.",
```

helm/cluster-aws/values.schema.json
```diff
@@ -196,13 +196,24 @@
                     "description": "Size of the volume mounted at `/var/log` on the worker nodes.",
                     "default": 30
                 },
+                "maxHealthyPercentage": {
+                    "type": "integer",
+                    "title": "Minimum percentage of instances that must remain healthy during node replacement.",
```
Gacko, line 201: Does this title also apply to this field? I feel like this is rather for `minHealthyPercentage`.

fiunchinho: Too much copy pasta :facepalm:

Gacko (approved):

PR #977, opened 2024-12-17
Explicitly set aws-node-termination-handler queue region so crash-loops are avoided, allowing faster startup

AndiDog:

Towards https://github.com/giantswarm/roadmap/issues/3802

Until now, NTH only started operating minutes after the cluster came up, or in unhealthy cluster conditions, even later. That could slow down ASG instance refreshes, node termination, etc. NTH only came up because the `AWS_REGION` environment variable is injected by IRSA.

The crash-looping message `FTL Unable to find the AWS region to process queue events.` goes away with this fix, but the pod still requires IRSA credentials injection to operate, so it may still take a few minutes to start up. But at least the error becomes clearer with this fix, and we avoid getting alerted.

### Checklist

- [x] Updated CHANGELOG.md.


fiunchinho (approved):

Gacko:
/run cluster-test-suites TARGET_SUITES=./providers/capa/china,./providers/capa/private

Gacko:
/run cluster-test-suites TARGET_SUITES=./providers/capa/china

AndiDog:
Tests are fine so far. Private cluster test failures are expected and fixed elsewhere (see https://github.com/giantswarm/aws-nth-bundle/pull/15).

PR #981, opened 2024-12-30
Explicitly set Ignition user data storage type to S3 bucket objects for machine pools

AndiDog:

This is a no-op right now as our fork still defaults to S3 storage, but needed because the [upstream CAPA feature PR](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5172) changes the default back to storage directly on the EC2 instance (for backward compatibility), so we must adapt configuration to explicitly choose the desired storage in S3.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests


Gacko:
Would it make sense to have this configurable?

AndiDog:
@Gacko No sense in making it configurable, given that it's a technical detail. Also, using storage on the EC2 instance limits us at 16 KB which turned out not enough for all our bootstrapping scripts, Teleport setup and stuff. We're bound to use external storage.

fiunchinho (approved):

PR #994, opened 2025-01-15
Chart: Reduce default etcd volume size to 50 GB.

Gacko:

Towards https://github.com/giantswarm/giantswarm/issues/32045.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


T-Kukawka:
what would be migration strategy for customers? adding that field in the values file?

Gacko:
What do you mean by migration strategy? The control plane nodes are getting replaced when upgrading Kubernetes anyway, so there's always a fresh etcd member joining the existing cluster and syncing data to a fresh volume.

In Vintage we kept the same etcd volumes, but this isn't the case for CAPI.

T-Kukawka:
my fear is that this is yet another breaking change for customers that they have to account for in their repository before upgrade

njuettner (approved):

T-Kukawka:
the target for this change is v30 correct?

Gacko:
Can you elaborate why customers would need to take actions in their repository?

T-Kukawka:
If they e.g, use 100GB now and want to keep it, they need to add it to the repository, if they have it configured as 100GB and want 50GB, they also have to adjust - so there is a change required in some cases. But maybe i am overreacting here.

Gacko:
Why would you want to use 100 GB? This is the volume size and we limited etcd to use only 8 GB by default and 16 GB by maximum if configured. So normally etcd would never exceed 16 GB of net usage. Even if fragmented and what not, it would probably not exceed 50 GB. We used to configure 100 GB because the GP2 volumes in AWS were scaling their IOPS with volume size. We never set 100 GB because any etcd member would actually use it. 😅

You can read more details in the referenced issue. 🙂

Gacko:

Gacko:
/run cluster-test-suites TARGET_SUITES=./providers/capa/upgrade

T-Kukawka:
if no real impact, regardless of the setting value, imma good 👍 thanks Marco

Gacko:
/run cluster-test-suites TARGET_SUITES=./providers/capa/upgrade

PR #991, opened 2025-01-14
Use reduced IAM permissions on worker nodes instance profile

AndiDog:
### What this PR does / why we need it

Towards https://github.com/giantswarm/roadmap/issues/3795, see https://github.com/giantswarm/capa-iam-operator/pull/369 which should be reviewed together with this PR

### Checklist

- [X] Updated CHANGELOG.md.

### Trigger E2E tests

Let's first roll out the operator before running tests – else it's a wasted test run.

PR #1005, opened 2025-01-23
Release v2.6.0

taylorbot:


AndiDog (approved):

AndiDog:
Tested via #977 and #991. The private cluster test failures will only be fixed with the aws-nth-bundle bump.

PR #1041, opened 2025-02-17
Chart: Update `cluster` to v2.0.0.

Gacko:


Gacko:

Gacko:
I'm going to enable the new apps in a separate PR.

Gacko:
There are breaking changes inside the v2, such as TLS ciphers, but these do not require changes to `cluster-aws`, especially because we added a condition to check on the Kubernetes version it's being used for.

So `cluster` v1.7.0 => v2.0.0 contains breaking changes to provider independent stuff, but does not directly require changes in at the consumer side (`cluster-aws`)

I'll enable the two new apps (`etcd-defrag` and `coredns-extensions`) in a second step. We do not have a workload cluster release including these, yet, so enabling them will break `cluster-test-suites` as long as there's not release including them and defining versions.

PR #1080, opened 2025-02-28
Add option to disable the node termination handler

fiunchinho:

Towards https://github.com/giantswarm/giantswarm/issues/32547

### Checklist

- [X] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


Gacko:
Weird, I was expecting us to also have some kind of `app.enabled` in the provider chart. But it seems this only exists in the cluster chart.

Gacko:
Ehm, if the `capa/standard` test is failing because AWS NTH Pods are restarting too often, then your change probably doesn't work, right? 😅

fiunchinho:
> Ehm, if the `capa/standard` test is failing because AWS NTH Pods are restarting too often, then your change probably doesn't work, right? 😅

This PR is adding an option to disable it. The default is to enable NTH, so it's expected that NTH is running.

Gacko:
Sure, I get that. But can we maybe also test the other way around by enabling the option for a moment?

fiunchinho:
> /run cluster-test-suites

Sure let's do it. [I've opened this](https://github.com/giantswarm/cluster-aws/pull/1082).

Gacko:
Thank you very much! I know, this is extra effort, but I'd like to prevent fixing this after it made its way into the releases. 😅

Gacko:
Tests are looking good, when disabling NTH: https://github.com/giantswarm/cluster-aws/pull/1082

paurosello (approved):

iuriaranda (approved):

PR #1131, opened 2025-03-26
feat(cert-manager): DNS01 challenge for CAPA

ssyno:
### Checklist

- [ ] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


helm/cluster-aws/templates/cert-manager-crossplane-resources.yaml
```diff
@@ -0,0 +1,47 @@
...
+  catalog: giantswarm
+  install:
+    timeout: "10m"
+  upgrade:
+    timeout: "10m"
+  kubeConfig:
+    inCluster: true # in management cluster context
+  name: cm-crossplane-resources
+  namespace: {{ $.Release.Namespace | quote }}
+  version: 0.0.3
```
Gacko, line 43: Please consume this from releases.

helm/cluster-aws/templates/cert-manager-crossplane-resources.yaml
```diff
@@ -0,0 +1,47 @@
...
+apiVersion: application.giantswarm.io/v1alpha1
+kind: App
+metadata:
+  labels:
+    app-operator.giantswarm.io/version: 0.0.0
+    {{- include "labels.common" $ | nindent 4 }}
+  name: {{ printf "%s-cm-crossplane-resources" (include "resource.default.name" $) | quote }}
+  namespace: {{ $.Release.Namespace | quote }}
+spec:
+  catalog: giantswarm
```
Gacko, line 34: Please consume this from releases.

Gacko (changes requested):
Please align the new app to others and consume releases where possible. Also you should add a check for if the app is actually contained in the release, otherwise it will break cluster test suites until there's a release out containing your app.

You can see the cluster chart for an example: https://github.com/giantswarm/cluster/blob/main/helm/cluster/templates/apps/apps.yaml

helm/cluster-aws/templates/cert-manager-crossplane-resources.yaml
ssyno: I did push the latest release to cluster-catalog and aligned everyhting I think

Gacko: I'm talking of Release CRs.

helm/cluster-aws/templates/cert-manager-crossplane-resources.yaml
Gacko: Same as for the version.

ssyno:
> Can we please call the app `cert-manager-crossplane-resources` where possible and at least when looking up stuff in the Release CR as this name will be customer-facing?

No we can't the name is very large

helm/cluster-aws/templates/_cert_manager_helmrelease_config.yaml
mcharriere, line 1: nitpick: cert-manager is not deployed via helmrelease. naming the file cert_manager_app_config.yaml or even cert_manager_config.yaml would be more appropriate

Gacko:
That's what I meant by "at least when looking up stuff in the Release CR".

I know resource names can be limited to 63 characters, but should not be the case for name in the app list of the Release CR.

ssyno:
> That's what I meant by "at least when looking up stuff in the Release CR".
> 
> I know resource names can be limited to 63 characters, but should not be the case for name in the app list of the Release CR.

should we use different names for the app and the helm?

helm/cluster-aws/templates/_cert_manager_helmrelease_config.yaml
ssyno: thanks for noting, renamed :)

Gacko:
You can use whatever name you want for the app and its configuration. It's just the app name that's being used for looking up the catalog and the version in the Release CR here: https://github.com/giantswarm/cluster-aws/pull/1131/files#diff-9de16bae3354ddc2ba48dc2fa04f137996f60e07c2e58e9f374dd677d3203e5fR34

This name will be the name listed in Release CRs and it would be nice if this is understandable for customers. Also it does not underlie the 63 characters limit.

ssyno:

ssyno:
/run cluster-test-suites TARGET_SUITES=./providers/capa/private

ssyno:

ssyno:
/run cluster-test-suites TARGET_SUITES=./providers/capa/private

ssyno:
/run cluster-test-suites TARGET_SUITES=./providers/capa/private

ssyno:
/run cluster-test-suites TARGET_SUITES=./providers/capa/private

ssyno:
/run cluster-test-suites TARGET_SUITES=./providers/capa/private

helm/cluster-aws/templates/cert_manager_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+      route53:
+        enabled: true
+        {{- $region := include "aws-region" . }}
+        region: {{ $region }}
+    http01:
+      enabled: false
+{{- else }}
+# Default empty configuration for non-private VPC mode
+{{- end }}
+{{- end }}
```
Gacko, line 27: Please add a newline at EOF.

helm/cluster-aws/values.schema.json
```diff
@@ -758,6 +759,12 @@
                             "title": "cert-manager",
                             "description": "Configuration of cert-manager. For all available values see https://github.com/giantswarm/cert-manager-app."
                         },
+                        "certManagerCrossplaneResources": {
+                            "$ref": "#/$defs/app",
+                            "type": "object",
+                            "title": "cert-manager Crossplane resources",
```
Gacko, line 765: I'd either go for `cert-manager-crossplane-resources` or `CertManager Crossplane Resources`.

helm/cluster-aws/templates/cert-manager-crossplane-resources-app.yaml
Gacko, line 1: Please make sure the Release CR supports this app, as in it includes information for this app, before rendering anything. Keep in mind that this version of `cluster-aws`, when released, might be used with an already existing release, which does not contain your app, yet.

This especially is the case for the daily `cluster-test-suites`. These use the latest released `cluster-aws` and the latest released CAPA Release. So as soon as your change makes it into a released version of `cluster-aws`, daily `cluster-test-suites` will fail.

Also, without this safeguard, we cannot include versions of `cluster-aws` which include this app in platform releases which are not yet meant to ship this app. This is the case for patch releases of existing CAPA Releases, like a future v30.1.1.

See the `cluster` chart for references: https://github.com/giantswarm/cluster/blob/main/helm/cluster/templates/apps/apps.yaml#L38-L44

For your particular use case, calling `is-in-release` should be enough, but all the other checks are also appreciated!

Gacko (changes requested):

helm/cluster-aws/values.schema.json
ssyno: I will go with `CertManager Crossplane Resources` to follow the same pattern as `Cilium Crossplane Resources`

helm/cluster-aws/templates/cert-manager-crossplane-resources-app.yaml
ssyno: Thanks for the detailed explanation! I added handling to check if the app exists in the releases before rendering anything. I modified both the `cert-manager-crossplane-resources` app installation and the `cert_manager_config` helm values to only apply when the cluster is private and the `cert-manager-crossplane-resources` app is included in the Release CR.

I hope this way the `cluster-test-suites` won't fail when using a newer `cluster-aws` with older `CAPA` Releases that don't include this app.

Are there any actions that I have to do to get the `cert-manager-crossplane-resources` app included in the upcoming releases?

Gacko: No, not really. You can of course put in a request in the respective `requests.yaml`, but as we are normally also requesting review of the involved teams, we should not miss adding your app!

Gacko (approved):
Thank you very much for your patience! I know it has been a lot of back and forth and confusion, but it looks pretty good now!

ssyno:
Opened [my request ](https://github.com/giantswarm/releases/pull/1649)
merging

PR #1402, opened 2025-09-04
Add cluster chart nodepool fields to the schema

fiunchinho:

When defining values for fields defined in the `cluster` chart schema but not in the `cluster-aws` schema, the schema validation for `cluster-aws` fails because those fields are not defined.

### Checklist

- [X] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Add `cluster` chart nodepool fields to the schema.
+
 ## [4.0.1] - 2025-09-02
 
 ### Changed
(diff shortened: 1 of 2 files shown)


iuriaranda (approved):
this means that we'll need to keep the nodepool schema in both places, which can be a pain to maintain (and to remember). Can't we just keep it in the aws chart (here)?

T-Kukawka (approved):

fiunchinho:
> this means that we'll need to keep the nodepool schema in both places, which can be a pain to maintain (and to remember). Can't we just keep it in the aws chart (here)?

Yeah, keeping them in both places sucks. But I think we need them in the `cluster` chart because we use those values in some templates in there. Also, we want to (eventually) use MachinePools on other providers like CAPZ.

PR #1478, opened 2025-10-07
Release v6.0.0

taylorbot:


Gacko (approved):

Gacko:
These won't work, @iuriaranda, because @fiunchinho changed how we are deploying Karpenter and this restructuring is taking effect in v33.0.0 first. So I _think_ at least Karpenter tests will fail, maybe even others.

iuriaranda:
Yep I just noticed 😅
I wanted to see if I could reproduce any of the test failures we've been seeing lately... I guess I'll create a new PR for it

PR #1758, opened 2026-02-11
feat(chart): Migrate Apps to HelmReleases.

fiunchinho:
## Summary

- Replace 6 AWS-specific App CR templates with Flux HelmRelease equivalents: `aws-nth-bundle`, `aws-pod-identity-webhook`, `aws-ebs-csi-driver-servicemonitors`, `irsa-servicemonitors`, `cert-manager-crossplane-resources`, `cilium-crossplane-resources`
- Add a pre-upgrade migration hook Job that safely removes chart-operator finalizers from AWS-specific Chart CRs and deletes them before the upgrade
- The `cluster` subchart has its own migration hook for the 20 generic apps (see companion PR)

### How the migration works
1. Two pre-upgrade hook Jobs run before `helm upgrade`:
   - The `cluster` subchart hook (weight -10) handles 20 generic apps
   - The `cluster-aws` hook (weight -5) handles 6 AWS-specific apps
2. For each migrated app, the hooks remove chart-operator finalizers from Chart CRs and delete them
3. This breaks the cascading deletion chain: when Helm deletes the old App CRs, chart-operator won't run `helm uninstall` because the Chart CRs are already gone
4. Helm creates the new HelmRelease CRs
5. Flux adopts the existing Helm releases via `helm upgrade`

### Related PR
- cluster chart: https://github.com/giantswarm/cluster/pull/780

## Test plan
- [x] Verify `helm template` renders HelmReleases instead of App CRs for all migrated apps (tested all 27 CI values files)
- [ ] Test upgrade on a staging cluster with existing App CRs
- [ ] Verify no app pods restart during the migration
- [ ] Verify HelmReleases show `Ready: True` after upgrade
- [ ] Verify Chart CRs are gone from `giantswarm` namespace after upgrade
- [ ] Test rollback scenario


Gacko:
/skip-ci Can not be tested here, has been tested there: https://github.com/giantswarm/releases/pull/2215.

PR #1896, opened 2026-04-27
Apply tags to subnets if they're defined by ID

### LINKED ISSUE #4083: Propagate subnet tags in cluster-aws even if a subnet is defined via ID
There was a case of wanting to edit subnets of a live cluster ([chat](https://gigantic.slack.com/archives/C6L8J93N0/p1756818914294129?thread_ts=1756720644.116399&cid=C6L8J93N0)). The `tags` field in `global.connectivity.subnets` right now doesn't end up in the `AWSCluster` object, so the desired state can't be reached via the manifest (but only manually). We should fix that since the source of truth should be in the cluster-aws values.

Watch out for vintage-migrated clusters – do they need to have their subnet tags added into the manifest before this change?

AndiDog:

Fixes https://github.com/giantswarm/roadmap/issues/4083

Tested manually (on top of cluster-aws v7.4.0 / CAPA Release v34.1.0): I created a workload cluster, whose subnet received the standard set of tags. Then I applied some `tags: { k: v, ... }` to a few public and private subnets. For one subnet, I overrode `giantswarm.io/role` to ensure the rendering logic doesn't output that tag twice. The additional tags got rendered and correctly reconciled (no tags were removed).

This is only relevant for CAPA-owned clusters, not for the "unmanaged" reocnciliation type in CAPA (adopt the network; we don't currently use that).

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


AndiDog:

T-Kukawka (approved):

AndiDog:

PR #2003, opened 2026-06-17
fix: nest NTH bundle values correctly

iuriaranda:

- The NTH values need to be under `upstream` otherwise they don't get propagated to the correct subchart.
- Tag AWSMachinePool ASG instances with `aws-node-termination-handler/managed: "true"` so NTH doesn't ignore interruption events from those instances. This worked without the tag before because NTH was configured to act upon events from all instances, but this could lead to some erratic behaviour or race-conditions with Karpenter.

### Checklist

- [x] Updated CHANGELOG.md.

### Trigger E2E tests

/run cluster-test-suites


helm/cluster-aws/templates/aws-nth-app.yaml
```diff
@@ -3,30 +3,32 @@
 {{- define "defaultAwsNodeTerminationHandlerHelmValues" }}
 awsNodeTerminationHandler:
   values:
-    awsRegion: {{ include "aws-region" $ | quote }}
+    global:
+      podSecurityStandards:
+        enforced: {{ .Values.global.podSecurityStandards.enforced }}
```
fiunchinho: I think this can be removed now

iuriaranda: the chart's default is `true`. You mean that we always go for that in all setups?

fiunchinho: I don't think that value is relevant anymore. If the chart is using it to gate any templates, that should be removed. We already removed that value from many charts https://gigantic.slack.com/archives/C05DCHUKTFH/p1771865061540749

iuriaranda: in this case, it toggles a Kyverno `PolicyException`, not a PSP. https://github.com/giantswarm/aws-nth-bundle/blob/main/helm/aws-node-termination-handler/templates/pss-exceptions.yaml

Does it mean that it needs to be always toggled on?

fiunchinho: I'm not sure, in all the repositories where I cleaned up this value, none had a PolicyException conditionally rendered depending on the value. I guess it always needs the PolicyException

iuriaranda:
@fiunchinho I've removed the PSS value and made some other changes, can you review again? 🙏

fiunchinho (approved):

PR #2006, opened 2026-06-18
fix: [v7.7 backport] add note-termination-handler tag on AWS machine pool

iuriaranda:

Without this, NTH just ignores any interruption event from those nodes

This was working before `aws-nth-bundle` v2.0.0 because we were using our own helm chart and it was forcing nth to ignore the tag.

But ideally NTH shouldn't ignore the tag, and it should only act on the instances that it manages, specially now that we have karpenter and it handles the interruption of its own instances.

### Checklist

- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


Gacko (approved):

iuriaranda:
closing. No need to backport it. Fix will be on the latest release

PR #2024, opened 2026-07-14
feat: add createIamRole opt-out for cert-manager crossplane role

paurosello:
## What

Adds a new `global.connectivity.certManager.createIamRole` toggle (default `true`) that lets customers opt out of provisioning the cert-manager IAM role via crossplane and **bring their own role**.

It layers on top of the existing `useCertManagerDnsChallenges` gate: when `createIamRole` is set to `false`, the `cert-manager-crossplane-resources` app (ConfigMap + App) is not rendered. Everything else — including the cert-manager ServiceAccount role-ARN annotation — is unchanged, so the customer is expected to provide a role with the conventional name `<clusterName>-cert-manager`.

Default behavior is unchanged (`createIamRole: true` reduces the guard to the original condition).

## Changes

- `templates/cert-manager-crossplane-resources-app.yaml` — add `createIamRole` to the guard condition
- `values.yaml` — `createIamRole: true` default
- `values.schema.json` — schema entry (normalized, passes `cluster-app` rule set)
- `README.md` — regenerated values docs
- `CHANGELOG.md` — Unreleased entry

## Verification

Rendered the template across the toggle matrix (in-release gate forced on):

| Case | Result |
|---|---|
| DNS on, `createIamRole` default (true) | App rendered ✓ |
| DNS on, `createIamRole=false` | App skipped ✓ |
| DNS off | App skipped (outer gate) ✓ |

`make validate-schema` passes (valid JSON Schema, normalized, valid per `cluster-app`).


T-Kukawka (approved):

AndiDog (approved):
Makes sense. We don't want to overcomplicate it for now and allow name override. I guess this needs a backport?

PR #2034, opened 2026-07-17
chore(release): v8.9.0

taylorbot:


iuriaranda (approved):

Gacko:
Please always run tests also in release PRs, as these trigger a different set of tests (change PRs only trigger Standard and maybe Upgrade IIRC, while release PRs also trigger Private, Cilium ENI Mode and China).

PR #2036, opened 2026-07-20
chore(release): v8.9.1

taylorbot:


glitchcrab:
/skip-ci no need to run tests, just a change to a taint value which was introduced in the previous release

AndiDog (approved):

PR #2039, opened 2026-07-22
feat: add per-node-pool localNvme.enabled option

luca-rui:
> **Draft — blocked on the `cluster` release.** See the checklist below.

## What

Adds a `localNvme.enabled` switch to the node pool schema, letting worker pools opt into local EC2 instance-store NVMe ephemeral storage for the kubelet directory (pod `emptyDir` + container runtime on fast node-local NVMe instead of the EBS lib/root volume).

The actual mechanism lives in the `cluster` chart (giantswarm/cluster#960). This PR only exposes the per-pool switch through the provider schema; `global.nodePools.<name>.localNvme.enabled` flows through to the `cluster` subchart via shared `global.*`, so no template wiring is needed here.

## Scope

Added to **both** node pool branches of `$defs.machinePool`:
- the standard (cluster-autoscaler) `machinePool` branch (`oneOf[1]`), and
- the **Karpenter** branch (`oneOf[0]`).

Earlier revisions of this PR scoped it to the non-Karpenter branch on the assumption that Karpenter nodes bootstrap via their own userData and would ignore worker `preKubeadmCommands`. **That assumption was wrong.** GS Karpenter nodes bootstrap via `KarpenterMachinePool` → a KubeadmConfig-generated bootstrap secret uploaded to S3 → an Ignition stub in `EC2NodeClass.userData` (`AMIFamily: Custom`) that merges from S3. The `cluster` chart's `preKubeadmCommands` therefore **do** run on Karpenter nodes, so `localNvme` is functional there — not a no-op.

## Validation (Karpenter, graveler)

Validated end-to-end on graveler with `cluster` shipping the mechanism and a `type: karpenter` pool opting in (via the escape-hatch `preKubeadmCommands` on the released chart, which is the exact path the flag drives once the dependency is bumped):

- Karpenter launched an instance-store node (`c5d.xlarge`) and the `cluster` script ran on it.
- The local NVMe disk backs `/var/lib/kubelet` — pod `emptyDir` lands on `/dev/nvme3n1` (the instance-store disk); node `ephemeral-storage` capacity == the NVMe disk, not the EBS root.
- The kubelet `evictionHard.nodefs.available: 10%` kubeadm patch is present on the running node (verified via `configz`).
- **Node-level `DiskPressure` eviction fires correctly**: a disk-filling pod was evicted at the threshold while an under-request pod and an idle pod survived.

Note: pin Karpenter pool `requirements` to instance-store families actually offered in the target region (e.g. `karpenter.k8s.aws/instance-family In [...]`); `r6id` is **not** offered in `eu-north-1`, for example.

## Why draft / blocked

The field is **inert** until the `cluster` dependency is bumped to a released version that ships the mechanism. That bump can't be written until `cluster` releases.

- [ ] giantswarm/cluster#960 merges
- [ ] `cluster` release cuts a version with the mechanism (expected v6.9.0)
- [ ] Bump the `cluster` dependency in `Chart.yaml` to that version (+ CHANGELOG `Chart: Update cluster to vX.Y.Z`)
- [ ] Mark ready for review

## Testing

`helm template` with `localNvme.enabled: true` on **both** a standard and a **Karpenter** node pool passes schema validation against the released `cluster` 6.8.0 subchart (renders cleanly, no-op as expected until the dependency bump). A bogus pool-level key is still rejected by the branch-level `additionalProperties: false`, and `make validate-schema` / `make generate-docs` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


luca-rui:
Reworked on top of `main` (10.1.0). The PR now carries the whole feature instead of only the schema change:

- `global.nodePools.<pool>.localNvme.enabled` for machinepool and karpenter pools. Enabling it runs
  `/opt/bin/setup-local-nvme.sh` before kubeadm on that pool only, through the new
  `preKubeadmCommandsTemplateName` hook from giantswarm/cluster#960 (`awsWorkersPreKubeadmCommands` in
  `_helpers.tpl`).
- The script ships like `wait-elb-dns.sh`: a key in the provider-specific-files Secret plus an entry in the
  worker `kubeadmConfig.files`. It finds instance-store NVMe disks through `/sys/block/*/device/model`, builds a
  RAID 0 when there are several, formats XFS and mounts it at `/var/lib/kubelet` with a systemd mount unit
  ordered after `var-lib.mount`, then writes the kubeadm patch `kubeletconfiguration2awsnvme+merge.yaml` with
  `evictionHard.nodefs.available: "10%"`. Nodes without instance store keep `/var/lib/kubelet` on the EBS lib
  volume.
- Karpenter pools with `localNvme.enabled` get the requirement `karpenter.k8s.aws/instance-local-nvme Exists`
  and the same `nodefs.available` threshold in `ec2NodeClass.kubelet.evictionHard`.
- `ci/test-local-nvme-values.yaml` with one enabled and one plain pool of each type.

Notes:

- The new `files:` entry changes every worker `KubeadmConfig`, so the upgrade rolls the worker nodes once. I did
  not bump the Secret suffix (`-5`): no existing file changed its content. Happy to bump if you prefer; that
  would also roll the control plane.
- An instance stop/start wipes the instance store and with it the kubelet configuration. Such a node has to be
  replaced. This is documented in the value description and in the script header.
- Stays draft: the cluster-aws CI is red until `cluster` releases giantswarm/cluster#960 and the dependency is bumped here.
- Board issue for the follow-ups: giantswarm/giantswarm#37692.

helm/cluster-aws/files/opt/bin/setup-local-nvme.sh
```diff
@@ -0,0 +1,117 @@
...
+# Preserve anything that already exists in the kubelet directory (normally nothing at this point).
+mkdir -p "${MOUNT_PATH}"
+if [ -n "$(ls -A "${MOUNT_PATH}")" ]; then
+  echo "Copying existing content of ${MOUNT_PATH} to the new filesystem"
+  tmp_mount="$(mktemp -d)"
+  mount -t xfs "${device}" "${tmp_mount}"
+  cp -a "${MOUNT_PATH}/." "${tmp_mount}/"
+  umount "${tmp_mount}"
+  rmdir "${tmp_mount}"
+fi
```
Gacko: That's not entirely true: `kubelet` is a systemd service and IIRC already running at this point. It's just waiting for a configuration created by `kubeadm`. So theoretically there could be some runtime data of `kubelet` in place and simply copying stuff would not deal with existing file handles.

luca-rui: Good point, thanks. `kubelet.service` is enabled in the image with `Restart=always` and exits at startup until `kubeadm` has written `/var/lib/kubelet/config.yaml`, so in practice it does not hold anything in `/var/lib/kubelet` yet, but the comment gave the wrong reason. The script now stops `kubelet.service` before it copies the directory and mounts the new filesystem, and relies on the `kubelet-start` phase of `kubeadm join` to start the kubelet again (it stops and starts the service anyway). Comments updated accordingly: 8a31528

luca-rui:
Pushed `002b749` with fixes from a review pass over `setup-local-nvme.sh`. The main finding: kubeadm re-runs the pre-kubeadm commands on every start of `kubeadm.service` until `kubeadm join` succeeds, including after a reboot, so the script has to be safe to run more than once. Changes:

- Before `mdadm --create` / `mkfs.xfs` the script stops RAID arrays that hold the instance-store disks, unmounts them and wipes the signatures, so a retry does not fail on busy members or an existing superblock. No `/etc/mdadm.conf` is written any more: udev assembles the array from the member superblocks, and the mount unit refers to the filesystem label.
- The kubelet patch is written before the mount starts, so the early exit on an existing mount point cannot skip it. It now also sets `nodefs.inodesFree: "5%"`, because a configured `evictionHard` map replaces the kubelet defaults.
- A `kubelet.service` drop-in with `RequiresMountsFor=/var/lib/kubelet` on nodes where the script moved the directory. If the instance store is gone (stop/start), the kubelet fails with a dependency error instead of starting on an empty EBS directory.
- Mode `0750` and the SELinux label are restored on the new filesystem root. `/sys/block` entries without a device node (NVMe multipath paths) are skipped.
- Karpenter pools with `localNvme.enabled` also exclude the `d` instance category (d3, d3en); their instance store is HDD.
- Schema, README and CHANGELOG describe the eviction thresholds and the stop/start behaviour, and correct one statement: enabling the option does not act on existing Karpenter nodes that already run on an instance type with instance store; they keep `/var/lib/kubelet` on the EBS lib volume until Karpenter replaces them. EBS-only nodes of that pool no longer match the requirement and get replaced through drift. The CHANGELOG states that the change rolls the worker nodes of `machinepool` node pools once.
- `ci/test-local-nvme-values.yaml` gains a Karpenter pool with the option enabled and default requirements, so both requirement paths render in CI.

Verified locally: `bash -n`, schemalint, both CI renders (the script command and the added requirements appear only in enabled pools; the Secret payload matches the file). Not verified on hardware yet: RAID re-assembly after a reboot on a two-disk type, and the drop-in behaviour after a stop/start. I plan a test-cluster run for both.

The Secret suffix stays at `-5`, as discussed above.

/run cluster-test-suites

AndiDog:
Superseding this PR with https://github.com/giantswarm/cluster-aws/pull/2091

